PR #5, opened 2015-09-11
Use delegate syntax

kjcpaas:
Use `delegate` for better readability in `MessageEvent` class.


sarsntony:
This breaks the gem on ruby 2.2.3, Rails 4.2.4. I can use the 1.0.4 gem without any issue, but any version after that breaks.

PR #12, opened 2015-10-12
Enhancements to keep track of more data

brpeterman:
### Bug fixes ###
Users were not being created correctly according to the responses I was getting from Discord. This is what they look like coming from the server:

```json
        "members": [
          {
            "user": {
              "username": "Oyerth",
              "id": "101794296201494528",
              "discriminator": "2305",
              "avatar": "f220a1a41c2e2b3408f309d54753d7ae"
            },
            "roles": [
              
            ],
            "mute": false,
            "joined_at": "2015-10-09T02:10:04.260000+00:00",
            "deaf": false
          },
```

And this is what the code was expecting:
```json
        "members": [
          {
            "username": "Oyerth",
            "id": "101794296201494528",
            "discriminator": "2305",
            "avatar": "f220a1a41c2e2b3408f309d54753d7ae",
            "roles": [
              
            ],
            "mute": false,
            "joined_at": "2015-10-09T02:10:04.260000+00:00",
            "deaf": false
          },
```

### Enhancements ###
I added some new members to some of the data classes:
- Server#channels
- User#status
- User#mute
- User#deaf
- Channel#topic

Additionally, Server#initialize now populates User#status based on the initial presence info.

I also added a default handler for the presence event which automatically updates user statuses.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -27,6 +27,8 @@ module Discordrb
 
       @channels = {}
       @users = {}
+
+      add_default_handlers
     end
 
     def run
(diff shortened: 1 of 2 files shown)


brpeterman:
So it looks like the concept of a "member" and a "user" are somewhat distinct. I'm going to have to make some changes... this change breaks events.

PR #14, opened 2015-10-14
Support for new events

brpeterman:
I made a bunch of changes here. See the first commit message for the overall summary.

Notably, I've added some structure to the relationship between servers, users, and channels and added the infrastructure to keep those relationships up to date as users and channels come and go.

Each server contains a set of users and channels. Users are always present in all text channels (this is an assumption that may not hold when we support permissions), and may be present in only one voice channel at a time. I think I've interpreted Discord correctly, but without an official API it's hard to be sure.

Let me know if you have questions, comments, or spot any bugs!

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -142,6 +142,19 @@ module Discordrb
     def channel_delete(attributes = {}, &block)
       register_event(ChannelDeleteEvent, attributes, block)
     end
+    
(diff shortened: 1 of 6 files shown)


brpeterman:
I added ten Array#find calls in this change, and I'm finding the need for more as I investigate implementing other events. It might be worthwhile to switch away from using arrays to store lists of users and channels, and start using hashes to make lookups by ID faster.

PR #35, opened 2016-01-10
Voice support

meew0:
Currently supports:
* Connection to a voice server
* Playback of files, URLs and arbitrary streams
* Pausing playback and continuing it again later
* Volume setting


meew0:
Apparently, houndci completely ignores my rubocop.yml...

PR #36, opened 2016-01-11
Fix bug on Voice::Encoder

purintai:
encode_file argument should be string. and file path maybe include space characters.


meew0:
:+1:

PR #43, opened 2016-02-10
Fix README to reflect proper testing procedures

kirillian:
There are no tests in this repository (yet). I recommend you remove the text trying to tell people how to run them until the tests are present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ You can find me (@meew0, ID 66237334693085184) on the unofficial Discord API ser
 
 ## Development
 
-After checking out the repo, run `bin/setup` to install dependencies. You can also run `bin/console` for an interactive prompt that will allow you to experiment.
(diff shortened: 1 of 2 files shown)


kirillian:
I'm sorry. I totally missed that spec folder. The rake false command fails as there is no rake task with the name 'false' (rake -T to see) and a quick inspection of the Rakefile confirms that. However, after looking again, it appears you are using rspec and rubocop without including the necessary dependencies in the Gemfile. I'll see if I can't get those working and make this PR reflect those changes instead.

kirillian:
Meant to post that comment earlier but posted it on the wrong PR.

kirillian:
You should now have a README that better reflects the commands necessary to run tests and the testing dependencies are now added to your gemspec file.

meew0:
Yeah, I added some tests once but never finished integrating them into the installation procedure, thanks for that

PR #44, opened 2016-02-10
Fix error resulting from accessing invite object as a hash instead of an object

kirillian:
This PR addresses an issue I ran into while implementing a bot using this API. ```bot#join``` throws up because the ```#invite``` method returns an object, rather than a hash. It appears this method is assuming a hash (maybe an implementation change at some point?). This is a relatively simple fix that should address that error.


kirillian:
I'm sorry. I totally missed that spec folder. The ```rake false``` command fails as there is no rake task with the name 'false' (```rake -T``` to see) and a quick inspection of the Rakefile confirms that. However, after looking again, it appears you are using rspec and rubocop without including the necessary dependencies in the Gemfile. I'll see if I can't get those working and make this PR reflect those changes instead.

PR #63, opened 2016-03-19
Disables rbnacl library if DISCORDRB_NONACL env var is set.

avengerx:
It does not just mute the message, it completely skips trying to load it.

### DIFF
--- a/lib/discordrb/voice/network.rb
+++ b/lib/discordrb/voice/network.rb
@@ -4,12 +4,14 @@ require 'socket'
 require 'json'
 
 begin
-  require 'rbnacl'
-  RBNACL_AVAILABLE = true
-rescue LoadError
-  puts "libsodium not available! You can continue to use discordrb as normal but voice support won't work.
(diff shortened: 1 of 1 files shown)


lib/discordrb/voice/network.rb
```diff
@@ -4,12 +4,14 @@
...
-  require 'rbnacl'
-  RBNACL_AVAILABLE = true
+  RBNACL_AVAILABLE = false
+  unless ENV::has_key?('DISCORDRB_NONACL')
+   require 'rbnacl'
+   RBNACL_AVAILABLE = true
+  end
 rescue LoadError
-  puts "libsodium not available! You can continue to use discordrb as normal but voice support won't work.
+   puts "libsodium not available! You can continue to use discordrb as normal but voice support won't work.
```
meew0: There seems to be an extra space here, can you remove it?

avengerx:
indentation on lines 9 and 10 were also wrong (1 space instead of 2 spaces)

PR #85, opened 2016-04-11
fix #split_message chopping off more than it got

AnhNhan:
When at the last element in the recursion, `ideal` became bigger than `msg`, since it contains all the newlines from the `join` operation.

This would cause 1) `msg[ideal.length..-1]` to become `nil`, causing future operations like `strip` to fail (and people probably not expecting `nil` in the returned value), and 2) `rest` containing superfluous / doubled newlines, causing some characters at the beginning of `rest` to be missing.

This is probably due to expecting `String#lines` to return lines with stripped newlines. Which for whatever reason is [not the case for Ruby](http://ruby-doc.org/core-2.3.0/String.html#method-i-each_line).

From my own limited test, this has fixed the issues I had experienced. I also added a test where reverting the change would reproduce the error.

### DIFF
--- a/lib/discordrb/data.rb
+++ b/lib/discordrb/data.rb
@@ -42,7 +42,7 @@ module Discordrb
     tri = [*0..(lines.length - 1)].map { |i| lines.combination(i + 1).first }
 
     # Join the individual elements together to get an array of strings with consecutively more lines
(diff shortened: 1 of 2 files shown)


AnhNhan:
Yay, I was fearing I actually broke something, but it's just coding style things.

PR #89, opened 2016-04-13
Add the ability to retrieve channels to light_bot

VxJasonxV:
These changes add a new API call to explicitly list channels for a given server_id, in addition to setting up the LightChannel structure to retrieve only the most critical channel attributes.

:warning: :fire: :skull: **not ready to be merged yet** :skull: :fire: :warning:

Hit List of changes still needing to be made:
- [ ] Bot Permissions need to be included at the channel level due to per-channel overrides
- [x] Documentation (:green_heart: RubyDoc)
- [x] Is a channel's topic critical? I think not, but feedback welcome.
- [x] Need the :white_check_mark: from Travis CI

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -288,6 +288,16 @@ module Discordrb::API
     )
   end
 
+  # List Channels for a Server
+  def list_channels(token, server_id)
+    request(
+      nil,
+      :get,
+      "#{api_base}/guides/#{server_id}/channels",
+      Authorization: token
+    )
+  end
+
   # Create a channel
   def create_channel(token, server_id, name, type)
     request(
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -293,7 +293,7 @@ module Discordrb::API
     request(
       nil,
       :get,
-      "#{api_base}/guides/#{server_id}/channels",
+      "#{api_base}/guilds/#{server_id}/channels",
       Authorization: token
     )
   end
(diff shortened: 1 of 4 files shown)


meew0:
Some comments:
 - Could you make a separate module for the channel attribute then include it in both kinds of channel classes to avoid code duplication? Like [`ServerAttributes`](https://github.com/meew0/discordrb/blob/master/lib/discordrb/data.rb#L1117).
- Maybe a utility method to get the channels directly from a `LightServer`?
- It seems the only thing Travis/RuboCop complains about is the commented-out stuff.

meew0:
Looks good now, will merge for the 2.1.0 release and once I've tested it

VxJasonxV:
There's no utility method yet to get channels from a LightServer yet, and the channel permission overrides don't work. Still working on those two bits.

meew0:
Yeah, just noticed too, I'd have probably noticed while testing though.

meew0:
Set to 2.2.0 so I can release 2.1.0, take your time

VxJasonxV:
LightBot can probably go away considering:
The API is more robust and faster
Webhooks imminently
RPC Impending
And laziness on my part ╮(. ❛ ᴗ ❛.)╭

greenbigfrog:
:sob:

meew0:
Not really; `LightBot` still has a valid use for OAuth applications that are not bots. It's true that you can use the `API` module directly instead, but `LightBot` makes it slightly easier by abstracting the token and the response parsing. Both webhooks and RPC serve different purposes (gateway- and authentication-less message sending and direct control of the Discord client, respectively) from `LightBot`.

VxJasonxV:
Fair. As far as me and the changes in this pull request are concerned;



I'll leave the branch around in case someone else wants to pull and continue on.

PR #92, opened 2016-04-15
Fixed the examples

snapcase:
Updated the examples

- bot initializers now use named parameters
  - a `prefix` needs to be specified
- the logic in some examples was set to break `if`the user id was present rather than `unless`
- the `eval` example would not allowed spaces unless `advanced_functionality` was enabled (disabled by default since 2.0.0)

### DIFF
--- a/examples/eval.rb
+++ b/examples/eval.rb
@@ -2,13 +2,13 @@
 
 require 'discordrb'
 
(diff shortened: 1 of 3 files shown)


meew0:
While you're at it, could you replace the email + PW authentication with token + app ID? That's the recommended way now.

green-big-frog:
Thanks for fixing! Somehow I had not watched out that the statements are correct as I switched it to `one` line...

PR #118, opened 2016-05-10
Updated the #prune Channel method to make use of bulk_delete

snapcase:
Updated to make use of bulk_delete

https://github.com/hammerandchisel/discord-api-docs/blob/master/docs/resources/CHANNEL.md#bulk-delete-messages--post-channelschanneliddocs_channelchannel-objectsmessagesbulk_delete

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -415,6 +415,18 @@ module Discordrb::API
     )
   end
 
+  # Delete messages in bulk
+  def bulk_delete(token, channel_id, messages = [])
+    request(
+      __method__,
+      :post,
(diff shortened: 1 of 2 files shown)


meew0:
Looks good, will test later

hlaaftana:
if it's over 100, you could loop over the multiples of 100 by the amount of  messages and wait 1 second in between to very easily avoid the 1/1 ratelimit

meew0:
Maybe, but it seems kinda pointless because you can just loop the call to `prune` (it gets new history every time)

meew0:
Apparently, bulk deleting one message won't work either, can you change the lower limit to 2? (Thanks @DinosParkour)

Issue #116, opened 2016-05-10
Channel exists with the given ID


SnazzyPine25:
I'm trying to see if a channel exists with the given ID
When I try `bot.channel()` with a non-existant ID it will put out a error.
When `bot.channel()` is given a non-existant ID it should output nil instead of putting out a error
Or have a function called `bot.channel?()`
The same thing should be preformed for `bot.user()` and finding servers

PR #124, opened 2016-05-16
Invalid Channel/User/Server ID returns nil

SnazzyPine25:


SnazzyPine25:
#116

meew0:
Can you make it rescue only the specific error that occurs? Otherwise it'll return nil when anything else happens (say, no network connection, or invalid token)

SnazzyPine25:
How could I do that exactly?

meew0:
Find out what error the method raises when called with an invalid ID, then update the rescue statement with only that error

SnazzyPine25:
`rescue 404`?

meew0:
No, it will raise `RestClient::Something`, which is what you should rescue

SnazzyPine25:
Alright.

PR #126, opened 2016-05-18
Awesome Logo

SnazzyPine25:
logo


meew0:
I spent a long time thinking about this logo, and while I certainly like the design, I don't think it really fits the project. In my opinion, it attracts too much attention without providing a lot of extra recognizability: a text just saying "DISCORDRB", no matter the colours, is a little generic. (Also, I don't feel like discordrb really needs a logo right now.)

If I ever think about having a logo in the future, I'll certainly take this into consideration.

PR #128, opened 2016-05-26
Idle Status/Streaming Status

SnazzyPine25:


lib/discordrb/bot.rb
```diff
@@ -489,6 +489,54 @@ def game=(name)
       name
     end
 
+    # Sets the currently online stream to the specified name and Twitch URL.
+    # @param name [String] The name of the stream to display.
+    # @return [String] The stream name that is being played now.
+    def stream=(name)
```
SnazzyPine25: Need help trying to add a argument to this. Trying to replace second arg with url.

meew0: You can't have two arguments on a writer, I'd recommend making it a regular method instead (like `def stream(name, url)`)

lib/discordrb/bot.rb
```diff
@@ -489,6 +489,54 @@ def game=(name)
       name
     end
 
+    # Sets the currently online stream to the specified name and Twitch URL.
+    # @param name [String] The name of the stream to display.
+    # @return [String] The stream name that is being played now.
+    def stream(name, url)
+      @game = name
+
+      data = {
```
meew0: Could you maybe make a generic `update_presence` method that handles presence events in general? The `stream`, `online`, `idle` and `game=` methods would then call that method to avoid duplicate code.

SnazzyPine25:
Should i use symbols to represent presence?

SnazzyPine25:
Should i keep the URL if any other type of presence change occurs?

SnazzyPine25:
@meew0 Am i supposed to merge the other update_presence or make the one i did with a different method name?

meew0:
Oh, I completely forgot there was already a method by that name... Can you rename it to something else? Maybe `send_presence_update` but feel free to come up with a more concise name

meew0:
Ah good, update_status is better. Looks good to me now, please fix the RuboCop errors though

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
         }
       }
-
       @ws.send(data.to_json)
+    end
+
+    # Sets the currently playing game to the specified game.
+    # @param name [String] The name of the game to be played.
+    # @return [String] The game that is being played now.
+    def game=(name)
```
meew0: This method doesn't actually set `@game`.

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
+      update_status(@idletime, @game, nil)
```
meew0: `@idletime` doesn't come from anywhere, so all methods that use it won't work (especially `idle`)

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
+      @idletime = idletime
+      @streamurl = url
+      type = url ? 1 : nil
       data = {
         op: Opcodes::PRESENCE,
         d: {
-          idle_since: nil,
-          game: name ? { name: name } : nil
+          idle_since: idletime,
+          game: name || url ? { name: name, url: url, type: type } : nil
```
meew0: This should use `@game` or `game` as the `name` variable no longer exists

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
     end
 
-    # Sets the currently playing game to the specified game.
-    # @param name [String] The name of the game to be played.
-    # @return [String] The game that is being played now.
-    def game=(name)
-      @game = name
-
+    # Updates presence status.
+    # @param idletime [Number] The floating point of a Time object.
```
meew0: This comment doesn't actually say what the parameter does

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
+    # @param game [String] The name of the game to be played. If url is not nil then it is the name of the stream to be streamed.
```
meew0: Both of these things are the same, so the second sentence is mostly redundant (maybe replace "to be played" with "to be played/streamed")

lib/discordrb/bot.rb
```diff
@@ -471,24 +471,53 @@ def parse_mention(mention)
...
+    # @param url [Number] The Twitch URL to display as a stream.
```
meew0: This comment should say that it can also be nil for no stream

SnazzyPine25:
Honestly I don't know how that's being a error

lib/discordrb/bot.rb
```diff
@@ -471,24 +473,54 @@ def parse_mention(mention)
...
+      update_status(nil, @game, @streamurl)
+    end
+    alias_method :on, :online
+
+    # Sets status to idle.
+    def idle
+      update_status((Time.now.to_f * 1000), @game, nil)
+    end
+    alias_method :away, :idle
+The
```
meew0: This

PR #130, opened 2016-06-01
Captions for files

SnazzyPine25:


lib/discordrb/api.rb
```diff
@@ -441,12 +441,13 @@ def acknowledge_message(token, channel_id, message_id)
   end
 
   # Send a file as a message to a channel
-  def send_file(token, channel_id, file)
+  def send_file(token, channel_id, file, caption = nil, tts = nil)
+	tts = caption && tts == nil ? false : tts
```
meew0: Instead of this hacky nil-checking stuff, could you make it use named keyword arguments?

SnazzyPine25: what do you mean?

meew0: `caption: nil, tts: false`, then users can use it like `send_file(file, caption: "some caption")` or `send_file(file, tts: true)` or whatever they like. Make sure to also apply this to the other methods

SnazzyPine25: Shold the docs stay the same?

lib/discordrb/api.rb
```diff
@@ -441,12 +441,13 @@ def acknowledge_message(token, channel_id, message_id)
   end
 
   # Send a file as a message to a channel
-  def send_file(token, channel_id, file)
+  def send_file(token, channel_id, file, caption: nil, tts: false)
+    tts = caption && tts.nil? ? false : tts
```
meew0: With named arguments, you don't need this line anymore

lib/discordrb/bot.rb
```diff
@@ -412,8 +412,10 @@ def send_message(channel_id, content, tts = false, server_id = nil)
     # @note This executes in a blocking way, so if you're sending long files, be wary of delays.
     # @param channel_id [Integer] The ID that identifies the channel to send something to.
     # @param file [File] The file that should be sent.
-    def send_file(channel_id, file)
-      response = API.send_file(token, channel_id, file)
+    # @param caption [string] The caption for the file.
+    # @param tts [true, false] Whether or not this file's caption should be sent using Discord text-to-speech.
+    def send_file(channel_id, file, caption: nil, tts: false)
+      response = API.send_file(token, channel_id, file, caption, tts)
```
meew0: Make sure to call the API method using named keyword arguments, not just have it accept them (otherwise you'll run into an ArgumentError)

meew0:
File captions are working fine, but `attach_file` doesn't seem to work at all - it just does nothing.

lib/discordrb/events/message.rb
```diff
@@ -44,6 +48,30 @@ def send_message(content)
...
+    # @return [Discordrb::Message] the message that was sent
+    def send_file(file, caption: nil)
+      @message.channel.send_file(file, caption: caption)
+    end
+
+    # Attaches a file to the message event and converts the message into
+    # a caption.
+    # @param file [File] The file to be attached
+    def attach_file(file)
+      @file = file.is_a(File) ? file : nil
```
meew0: Should be `is_a?`

meew0:
Still doesn't seem to work; same symptoms as before - attach_file does nothing

lib/discordrb/events/message.rb
```diff
@@ -44,6 +48,30 @@ def send_message(content)
...
+    # @return [Discordrb::Message] the message that was sent
+    def send_file(file, caption: nil)
+      @message.channel.send_file(file, caption: caption)
+    end
+
+    # Attaches a file to the message event and converts the message into
+    # a caption.
+    # @param file [File] The file to be attached
+    def attach_file(file)
+      @file = file.is_a?(File) ? file : nil
```
meew0: Also, could you make this raise an exception instead of just setting the file to nil?

SnazzyPine25:
What's making conflict?

lib/discordrb/events/message.rb
```diff
@@ -44,6 +48,30 @@ def send_message(content)
...
+    # @return [Discordrb::Message] the message that was sent
+    def send_file(file, caption: nil)
+      @message.channel.send_file(file, caption: caption)
+    end
+
+    # Attaches a file to the message event and converts the message into
+    # a caption.
+    # @param file [File] The file to be attached
+    def attach_file(file)
+      file.is_a?(File) ? @file = file : raise ArgumentError, "Argument is not a file!"
```
meew0: Can you turn this into an if/else statement? The ternary operator isn't really meant for this

lib/discordrb/events/message.rb
```diff
@@ -62,7 +62,7 @@ def send_file(file, caption: nil)
     # a caption.
     # @param file [File] The file to be attached
     def attach_file(file)
-      file.is_a?(File) ? @file = file : raise ArgumentError, "Argument is not a file!"
+      if file.is_a?(File); @file = file else raise ArgumentError, "Argument is not a file!"; end
```
hlaaftana: This looks kinda ugly lol

meew0: Yeah, just use newlines

lib/discordrb/events/message.rb
```diff
@@ -135,7 +164,11 @@ def matches?(event)
 
     # @see EventHandler#after_call
     def after_call(event)
-      event.send_message(event.saved_message) unless event.saved_message.empty?
+      if @file.nil?
```
meew0: This is the problem with the PR - `@file` only exists in the *event*, not in the *event handler*. Use `event.file` and it should work

PR #137, opened 2016-06-10
Added #playing? And #length for VoiceBot

SnazzyPine25:


meew0:
What exactly would `length` be useful for? It's an internal detail for which I can't imagine a valid use case.

SnazzyPine25:
Stream time and length can determine a good percentage of the song that's played, that's all I got. Honestly I just saw length being used and jumped to it

meew0:
The length is slightly misleadingly named, it's not the length of the song but rather the length the VoiceBot has to wait after having sent one packet. This is a detail that is useless to expose to the public.

SnazzyPine25:
Alright, should I delete that?

meew0:
Yes please, as it will likely mislead people looking at the docs

lib/discordrb/voice/voice_bot.rb
```diff
@@ -120,6 +121,14 @@ def pause
       @paused = true
     end
 
+    # @see #playing=
```
meew0: What method is this referencing? There is no `playing=` method in `VoiceBot`

PR #143, opened 2016-06-17
Playing Event Streaming Support

barkerja:
Adds support for (the not yet documented) `type` and `url` in the presence payload. This indicates if a member is streaming (`type` indicating which service: 0 no stream, 1 twitch; and a `url` provided to the location of the stream).

### DIFF
--- a/lib/discordrb/container.rb
+++ b/lib/discordrb/container.rb
@@ -123,6 +123,7 @@ module Discordrb
     # @param attributes [Hash] The event's attributes.
     # @option attributes [String, Integer, User] :from Matches the user whose playing game changes.
     # @option attributes [String] :game Matches the game the user is now playing.
+    # @option attributes [Integer] :type Matches the type of game object (0 game, 1 Twitch stream)
     # @yield The block is executed when the event is raised.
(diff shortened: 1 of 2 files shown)


meew0:
From a quick look, this seems to be a duplicate of #128. I'll take a more thorough look later, maybe there is something in your PR that the other doesn't

barkerja:
> From a quick look, this seems to be a duplicate of #128. I'll take a more thorough look later, maybe there is something in your PR that the other doesn't

Oops, missed that one! After taking a quick look at #128, their PR appears to only update the type/url on Discord but doesn't provide a way to access that data through discordrb in any way when received by Discord.

Maybe we merge the two PRs?

meew0:
Seems good, I'll take care of that when I review and merge all the PRs (either today or tomorrow)

SnazzyPine25:
Nice!

SnazzyPine25:
@barkerja can you also add `:streaming` to status?

meew0:
Streaming is not a separate status though - a user can both be streaming and `:online` and streaming and `:idle`.

Issue #141, opened 2016-06-15
Spam of "An existing connection was forcibly closed by the remote host"


hlaaftana:
```
[ERROR : gateway @ 2016-06-14 08:07:42.851] Gateway error: An existing connection was forcibly closed by the remote host.
[ERROR : gateway @ 2016-06-14 08:07:42.852] Gateway error: An existing connection was forcibly closed by the remote host.
[ERROR : gateway @ 2016-06-14 08:07:42.852] Gateway error: An existing connection was forcibly closed by the remote host.
[ERROR : gateway @ 2016-06-14 08:07:42.853] Gateway error: An existing connection was forcibly closed by the remote host.
[ERROR : gateway @ 2016-06-14 08:07:42.853] Gateway error: An existing connection was forcibly closed by the remote host.
```

Spams this when i cut my internet for a second / close my laptop's lid

wakeuphate:
Maybe worth nothing that I also get this, however it tends to happen at random on my Windows VPS.  Could just be a dodgy connection as per the original issue, but i've not noticed the connection drop when this happens.

SnazzyPine25:
I'm guessing it's saying that the computer itself stopped hosting it. You say it happens when the Internet cuts out for a second. It is a rare occasion. Probably a sensitive connection too.

hlaaftana:
Not really helping, the issue here is that it spams the error message instead of closing the websocket

meew0:
Maybe I'll just add a gateway error handler that reconnects whenever an error occurs, as I can't really fix this specific problem

hlaaftana:
Or just try to close the connection and interrupt the websocket thread?

PR #148, opened 2016-06-18
Fix gateway error spam

meew0:
#141. This needs testing by those who encounter the error so I'm making this a PR. @hlaaftana @wakeuphate


meew0:
Confirmed to work on Discord

Issue #144, opened 2016-06-17
Command chains don't work with Event#<< and other similar methods


unleashy:
Simply returning at the command block works though.

PR #149, opened 2016-06-18
Fix << for command chains

meew0:
#144. I tested it already but it would be nice if you @unleashy could test it as well for your specific situation.


unleashy:
It's working! :smile:

PR #154, opened 2016-06-19
Add send_message and the alike to MessageIDEvent

hlaaftana:
For #121


meew0:
Could you just make all that a separate module and include it in both classes to avoid code duplication like that?

hlaaftana:
Oh, sure

PR #155, opened 2016-06-19
Add autocorrect for rubocop

hlaaftana:
Trailing whitespace & other jazz are annoying to fix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ rvm:
 before_install: gem install bundler -v 1.10.6
 script:
   - bundle exec rspec spec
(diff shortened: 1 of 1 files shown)


meew0:
This will only fix it on the codebase currently running in the Travis container. It will have no effect on PRs or anything like that.

hlaaftana:
Oh right

Should I close then?

meew0:
I'll do it. As a note, you can always run `rubocop` locally yourself before committing so you don't have to wait for Travis each time. There's even an overcommit config file in the repo if you want to have git prevent you from committing unless rubocop passes (as I have it set up for my local copy)

PR #178, opened 2016-08-09
Add :required_roles attribute to CommandBot

PoVa:


PoVa:
Note: changes to `Member#role?` make it so that you can't pass a string as id to the method. Maybe I should create a separate method `Member#role_by_name?` (I'm bad at naming things)?

meew0:
I would drop being able to check roles by strings altogether, because it's a possible security issue - consider the following situation:

On the server there are three roles, "Moderator", "Trusted Member" and "Member". The "Moderator" role has the obvious moderation permissions. "Trusted Member" has `manage_roles` to be able to assign the "Member" roles to users on the server they see fit to receive it. The "Member" role has no particular permissions, it just serves as a designation that this person is active on the server.

Now there is a bot on the server that has the "Moderator" role, and its purpose is to ease some aspect of moderation (say, softbans). To prevent everyone from using it, the author has implemented a check that works like this:
```rb
if event.author.role? "Moderator"
  # Command user is a moderator! Do the task
  # [...]
end
```
(Here I used `role?` as an example, the same issue would occur with `required_roles`.) 

Now a "Trusted Member" turns out not to be that trustworthy and, using his `manage_roles` permission, he creates a new role also called "Moderator". According to Discord hierarchy roles, he can give this role to himself (or anyone else), but not the other "Moderator" role as it is above "Trusted Member" in hierarchy.

You can see where this is going: As the bot now considers the "Trusted Member" to have the "Moderator" role, the "Trusted Member" can use moderation commands as he pleases. If the bot author had used the role ID instead, this would never be a problem.

There is of course considerable utility benefit in using `role?` with a string, and preventing this doesn't get rid of the problem (people can still do the string check themselves), but having this here does make it easier. I'd prefer if this were not the case.

megumisonoda:
I'd hate to see this left unmerged just because it makes checking via string easier. I would propose that it be merged, and add a warning about its usage in the docs. That way, people know the potential risks, but if they'd like to implement their own sanity and safety checks on environment, or they trust their staff enough, they can still use it and have the huge utility benefits.

I myself would love to have this, personally: My bot runs on a single server and I make several assumptions in the code about that already (plus sanity+safety checks).

greenbigfrog:
(You could PR into @PoVa 's fork. Might speed it up)

meew0:
A doc warning could work but I'd still prefer to just remove role checking by string. Either way, @PoVa (or someone PRing into his fork as @greenbigfrog said) would have to change something.

Issue #182, opened 2016-08-14
voice_channel#users crashes after a few calls


Apexal:
Using the latest release, (v 2.1.3) `channel#users` works fine for both text and voice channels.
When I use the latest GitHub version, **after one or two calls (where it works correctly)** I get this error:

```
[ERROR : et-4 @ 2016-08-12 17:19:51.664] Exception: #<NoMethodError: undefined method `id' for nil:NilClass>
[ERROR : et-4 @ 2016-08-12 17:19:51.665] /home/ontrac/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/discordrb-2.1.3/lib/discordrb/data.rb:1083:in `block in users'
[ERROR : et-4 @ 2016-08-12 17:19:51.665] /home/ontrac/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/discordrb-2.1.3/lib/discordrb/data.rb:1083:in `each'
[ERROR : et-4 @ 2016-08-12 17:19:51.665] /home/ontrac/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/discordrb-2.1.3/lib/discordrb/data.rb:1083:in `map'
[ERROR : et-4 @ 2016-08-12 17:19:51.666] /home/ontrac/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/discordrb-2.1.3/lib/discordrb/data.rb:1083:in `users'
...
```

v2.1.3 version of `#users`: **WORKING**
```ruby
def users
  if @type == 'text'
    @server.members.select { |u| u.status != :offline }
  else
    @server.members.select do |user|
      user.voice_channel.id == @id if user.voice_channel
    end
  end
end
```

Latest GitHub version of `#users`: **NOT WORKING**
```ruby
def users
  if @type == 'text'
    @server.online_members(include_idle: true).select { |u| u.status != :offline }
  else
    @server.voice_states.map { |id, voice_state| @server.member(id) if voice_state.voice_channel.id == @id }.compact
  end
end
```

meew0:
I see the problem - it treats voice states without a channel as if they had one. Adding a nil check to inside the block should fix it.

Apexal:
And how does `text_channel#users` work? It looks like it just selects **all** non-idle members without checking if they have perms in the text-channel.

meew0:
[...]

PR #183, opened 2016-08-15
Fix voice_state issue with channel#users (Issue #182)

Apexal:
Added a nil check as suggested


PoVa:
Use `&&` instead of `and` to follow [ruby style guidelines](https://github.com/bbatsov/ruby-style-guide#no-and-or-or).

greenbigfrog:
#182 (Titles don't reference...)

Issue #169, opened 2016-07-07
Today i encountered this | stack level too deep | Segmentation Fault


Yixn:
```
[ERROR : gateway @ 2016-07-07 15:22:12.753] Exception: #<SystemStackError: stack level too deep>
[ERROR : gateway @ 2016-07-07 15:22:12.759] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:886:in `maybe_convert_extension'
[ERROR : gateway @ 2016-07-07 15:22:12.759] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:618:in `block (2 levels) in stringify_headers'
[ERROR : gateway @ 2016-07-07 15:22:12.760] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:617:in `map'
[ERROR : gateway @ 2016-07-07 15:22:12.760] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:617:in `block in stringify_headers'
[ERROR : gateway @ 2016-07-07 15:22:12.761] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:604:in `each'
[ERROR : gateway @ 2016-07-07 15:22:12.761] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:604:in `inject'
[ERROR : gateway @ 2016-07-07 15:22:12.762] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:604:in `stringify_headers'
[ERROR : gateway @ 2016-07-07 15:22:12.762] /root/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rest-client-2.0.0/lib/restclient/request.rb:434:in `make_headers'
[...]

Issue #170, opened 2016-07-27
Stack Overflow Error


jabarri1:
I get different sets of errors running locally and through VPS.

Local: https://gist.github.com/jabarri1/c7fa49072039d1ffa4ce577b710bb40b

VPS: https://gist.github.com/jabarri1/e78a0ab4e5df46682afe6b752342739b

PR #191, opened 2016-08-23
bug; infinite loop in server initialization if server owner is offline

stoodfarback:


stoodfarback:
When connecting to discord and initializing the `Server` objects, if the owner of a server is not listed with the members (this happens if the server has more than 100 members and the owner is offline), the code gets stuck in a loop of initializing `Server` -> fetching server owner info -> initializing `Server`.

This fixes it by passing the `Server` object along to the member lookup, so it can use that instead of trying to find it in the cache.

Likely fixes #169 and #170.

meew0:
Thanks for debugging this; it's one of the more pressing issues with 2.1.3 and I had no idea it had to do with the server owner. I'll make a couple comments to code style but functionality wise this seems good.

lib/discordrb/cache.rb
```diff
@@ -85,15 +85,16 @@ def server(id)
       @servers[id] = server
     end
 
-    # Gets a member by both IDs
-    # @param server_id [Integer] The server ID for which a member should be resolved
+    # Gets a member by both IDs, or `Server` and user ID.
+    # @param server_or_id [Integer] The `Server` or server ID for which a member should be resolved
```
greenbigfrog: This doesn't expect an `[Intger]` anymore

meew0: Here, the type parameter should now be `[Integer, Server]`.

stoodfarback: Ah, missed that. Updated pr.

lib/discordrb/cache.rb
```diff
@@ -85,15 +85,16 @@ def server(id)
       @servers[id] = server
     end
 
-    # Gets a member by both IDs
-    # @param server_id [Integer] The server ID for which a member should be resolved
+    # Gets a member by both IDs, or `Server` and user ID.
+    # @param server_or_id [Server, Integer] The `Server` or server ID for which a member should be resolved
     # @param user_id [Integer] The ID of the user that should be resolved
     # @return [Member, nil] The member identified by the IDs, or `nil` if none could be found
```
greenbigfrog: could you please decapitalize the return if you're at it? looks weird in the cods if it's not... (`The` -> `the`)

stoodfarback: Sorry, I don't understand what you mean.

PR #195, opened 2016-08-26
Add alias dm (direct_message) for pm (private_message)

greenbigfrog:
Technically pm's are dm's in discord...


greenbigfrog:
According to github search I should have added all... you didn't answer yet if I should do `DM/PM` everywhere in docs f.e.

meew0:
I'd keep it as PM everywhere in docs since the methods are called that (the aliases only show up marginally).

greenbigfrog:
So you should be able to merge this...

PR #197, opened 2016-08-28
Add support for server's channel endpoint

depl0y:
This change adds support for the channels endpoint for a server. The authorization token and server_id need to be supplied for this method.

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -190,6 +190,15 @@ module Discordrb::API
     )
   end
 
+  def channel_invites(token, channel_id)
+    request(
+      __method__,
+      :get,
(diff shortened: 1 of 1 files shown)


lib/discordrb/api.rb
```diff
@@ -180,6 +180,25 @@ def bans(token, server_id)
...
+  # Get a server's channels list
+  def channels(token, server_id)
+    request(
+      __method__,
+      :get,
+      "#{api_base}/guilds/#{server_id}/channels",
+      Authorization: token
+    )
+  end
+
```
meew0: Could you add an one-line comment here similar to the other methods?

depl0y: I committed a change containing the comment.

PR #199, opened 2016-08-30
Add :no_permission_message attribute to CommandBot

PoVa:


greenbigfrog:
The issue like it's ATM, is that a user get's displayed the `Bot#invite_url`, but adding it again won't fix anything at all, since it has the same, when not even less, permissions...

What you could do is probably add something like `Bot#invite_url(permissionbits)`...

greenbigfrog:
reopened in #200

PR #201, opened 2016-08-30
Added a parameters option for command def

FormalHellhound:
I added a parameters option for the command def, so that the built in !help command can have a little more detail.

You would assign an array to your parameters attribute like so `parameters: ["Thing1, Thing2"]` and then when the !help command is run it will display the parameters if they exist like so:

Parameters:
        `Thing 1`
        `Thing 2`

This would be helpful for commands that accept multiple different parameters.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -127,6 +127,8 @@ module Discordrb::Commands
           usage = command.attributes[:usage]
           result = "**`#{command_name}`**: #{desc}"
           result += "\nUsage: `#{usage}`" if usage
+          result += "\nAccepted Parameters:" if parameters
+          parameters.each { |p| result += "\n    `#{p}`"} if parameters
           result
         else
           available_commands = @commands.values.reject { |c| !c.attributes[:help_available] }
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -125,6 +125,7 @@ module Discordrb::Commands
           return "The command `#{command_name}` does not exist!" unless command
           desc = command.attributes[:description] || '*No description available*'
           usage = command.attributes[:usage]
+          parameters = command.attributes[:parameters]
           result = "**`#{command_name}`**: #{desc}"
           result += "\nUsage: `#{usage}`" if usage
           result += "\nAccepted Parameters:" if parameters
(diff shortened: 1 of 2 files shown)


meew0:
This is kinda what the `usage` option is supposed to be for - a field in which you can list parameters and how they should be used in a free-form format. What advantages does an array of parameters have?

FormalHellhound:
Well for example we have one command that posts a reaction face picture based on the parameter given like: `!r smug`, and as we continue to add more there could end up being more than 20, I took usage to be where I would show how a command is formatted, and how many arguments are required like `usage: "!r <parameter>` or if a command took more than one parameter at a time like `usage: !r <parameter1> <parameter2>`

I thought that taking an array of possible parameters, and formatting it so that they are in a code block with one parameter per line makes it much easier to read than `Usage: !r <smug/angry/happy/no/etc/etc/etc/etc/etc/etc> <etc/etc/etc/etc/etc/etc/etc>`

greenbigfrog:
Doing it on multiple lines, creates way more spam... Maybe then dm the help message.

meew0:
That's certainly a valid point.

meew0:
@greenbigfrog discordrb already manages DMing help messages, no need to do that

lib/discordrb/commands/command_bot.rb
```diff
@@ -125,8 +125,11 @@ def initialize(attributes = {})
           return "The command `#{command_name}` does not exist!" unless command
           desc = command.attributes[:description] || '*No description available*'
           usage = command.attributes[:usage]
+          parameters = command.attributes[:parameters]
           result = "**`#{command_name}`**: #{desc}"
           result += "\nUsage: `#{usage}`" if usage
+          result += "\nAccepted Parameters:" if parameters
```
meew0: In addition to what Rubocop already remarked, these two lines can be turned into one ```rb 
if parameters
  ...
end
```
statement.

**Comment by FormalHellhound** [comment]
I may take a look and see if I can do this with a hash instead, so if a command does take multiple arguments at the same time, and parameter 2 depends on what parameter 1 is `!command <parameter1> <parameter2>`, it can be shown more clearly with indented code blocks like:

Parameters:
```
  parameter1:
    etc
      parameter2:
        etc
        etc
```

meew0:
To me, that just seems unnecessarily complicated; I can see how a simple list of possible parameters can be useful but I would put clarifying information in the usage section.

PR #207, opened 2016-09-03
Use code blocks for examples/eval.rb

SnazzyPine25:


meew0:
This makes no sense to do, because the result of an `eval` instruction doesn't have to be Ruby code - it could be an arbitrary string. Likewise, if you print the exception like that, it will simply print the message, which is not Ruby code either.

PR #208, opened 2016-09-04
Reintroduce Bot#stop

megumisonoda:
Also fix `Gateway#stop` in the process.

`@ws.close` was still in use from pre-`gateway_rewrite` code, and was replaced with the new `close`.

In addition, the `connect_loop` method was continually setting `@should_reconnect = true` which caused errors, so this was moved outside the loop block.

Finally, `handle_close` assumed that the websocket would always close due to an error. Due to the potential for this behavior to be misleading, it now does nothing if `e` is `nil`. I suspect that you may prefer a different way of handling this.

Regression from the previous `Bot#stop` implementation: It is now impossible to use it to catch `SIGINT` and `SIGTERM`, since `close` relies on `.synchronize`, which cannot be called from a `trap` context.

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -696,7 +696,11 @@ module Discordrb
       # This needs to be synchronised with the getc mutex, so the notification, and especially the actual
       # close afterwards, don't coincide with the main loop reading something from the SSL socket.
       # This would cause a segfault due to (I suspect) Ruby bug #12292: https://bugs.ruby-lang.org/issues/12292
-      @getc_mutex.synchronize { @closed = true } unless no_sync
+      if no_sync
+        @closed = true
+      else
+        @getc_mutex.synchronize { @closed = true }
+      end
 
       # Close the socket if possible
       @socket.close if @socket
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -162,12 +162,12 @@ module Discordrb
     end
 
     # Stops the bot gracefully, disconnecting the websocket without immediately killing the thread. This means that
-    # Discord is immediately aware of the closed connetion and makes the bot appear offline instantly.
+    # Discord is immediately aware of the closed connection and makes the bot appear offline instantly.
     #
     # If this method doesn't work or you're looking for something more drastic, use {#kill} instead.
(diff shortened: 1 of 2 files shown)


lib/discordrb/gateway.rb
```diff
@@ -638,17 +638,19 @@ def handle_internal_close(e)
...
-        LOGGER.error('Websocket close frame received!')
-        LOGGER.error("Code: #{e.code}")
-        LOGGER.error("Message: #{e.data}")
-      elsif e.is_a? Exception
-        # Log the exception
-        LOGGER.error('The websocket connection has closed due to an error!')
-        LOGGER.log_exception(e)
-      else
-        LOGGER.error("The websocket connection has closed due to an unspecified error: #{e.inspect}")
+      unless e.nil?
```
meew0: In my opinion this method should *always* log what it does, so if for some reason the websocket closes with something I didn't account for, it could be caught and added. How about changing the `else` case to an `info` call, and just saying `"The websocket connection has closed. (#{e.inspect})"`?

meew0:
As for the synchronize thing, how about having an argument to `close` (and `Gateway`/`Bot#stop` in propagation) that is false by default and, if true, doesn't do any synchronisation? That way if someone wanted to call it from a trap, they can again now, with the default behaviour still being the safe one.

lib/discordrb/gateway.rb
megumisonoda: Will do.

megumisonoda:
That's a good idea. Expect new changes tomorrow, most likely.

lib/discordrb/gateway.rb
```diff
@@ -696,7 +696,7 @@ def close
       # This needs to be synchronised with the getc mutex, so the notification, and especially the actual
       # close afterwards, don't coincide with the main loop reading something from the SSL socket.
       # This would cause a segfault due to (I suspect) Ruby bug #12292: https://bugs.ruby-lang.org/issues/12292
-      @getc_mutex.synchronize { @closed = true }
+      @getc_mutex.synchronize { @closed = true } unless no_sync
```
meew0: This looks wrong; if `no_sync` is disabled it won't set `@closed` at all.

PR #213, opened 2016-09-06
Various method parameter code cleanup

Daniel-Worrall:
Replaced various parameter default syntax for consistency

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -530,7 +530,7 @@ module Discordrb::API
   end
 
   # Send a file as a message to a channel
-  def send_file(token, channel_id, file, caption: nil, tts: false)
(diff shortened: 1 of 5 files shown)


meew0:
These syntaxes do different things, if you use `=` you can only use the arguments in one order but don't have to specify the name, with `:` you can put them in any order but have to specify the name.

```rb
def foo(a, b = 1, c = 2)
  p a, b, c
end

foo(3) # => 3, 1, 2
foo(3, 2) # => 3, 2, 2
foo(3, 2, 3) # => 3, 2, 3

def bar(a, b: 1, c: 2)
  p a, b, c
end

bar(3) # => 3, 1, 2
bar(3, c: 4) # => 3, 1, 4
bar(3, c: 1, b: 2) # => 3, 2, 1
bar(3, 2, 1) # invalid
```

PR #217, opened 2016-09-11
Add support for multiple streams

PoVa:


lib/discordrb/logger.rb
```diff
@@ -9,16 +9,17 @@ class Logger
     # @return [true, false] whether this logger is in extra-fancy mode!
     attr_writer :fancy
 
-    # @return [IO, #puts & #flush] the stream the logger should write to.
-    attr_accessor :stream
+    # @return [<Array>IO, <Array>#puts & <Array>#flush] the streams the logger should write to.
```
meew0: This type parameter should say `Array<IO>, Array<#puts & #flush>`

lib/discordrb/logger.rb
```diff
@@ -9,16 +9,17 @@ class Logger
...
+    attr_accessor :streams
 
     # Creates a new logger.
     # @param fancy [true, false] Whether this logger uses fancy mode (ANSI escape codes to make the output colourful)
-    def initialize(fancy = false)
+    # @param streams [<Array>IO, <Array>#puts & <Array>#flush] the streams the logger should write to.
```
meew0: Same here

lib/discordrb/logger.rb
```diff
@@ -9,16 +9,17 @@ class Logger
...
+    def initialize(fancy = false, streams = nil)
```
meew0: You can just set the default value to `[STDOUT]` and simplify the lower expression to `@streams = streams`

PR #228, opened 2016-09-19
Fix for private channel issue

heimidal:
Just needed to convert the contents to JSON and add the content type.


greenbigfrog:
nice catch 👍

PR #230, opened 2016-09-21
Reintroduces heartbeat events that were lost in the gateway update

### LINKED ISSUE #229: Heartbeat events are no longer raised
Since the gateway rewrite that changed the way heartbeats were handling, the raised events were removed in the process.

Daniel-Worrall:
Fixes #229

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -353,6 +353,7 @@ module Discordrb
             # suspended (e.g. after op7)
             if (@session && !@session.suspended?) || !@session
               sleep @heartbeat_interval
+              @bot.raise_heartbeat_event(Discordrb::Events::HeartbeatEvent.new(self))
               heartbeat
             else
               sleep 1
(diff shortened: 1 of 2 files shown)


lib/discordrb/gateway.rb
```diff
@@ -353,6 +353,7 @@ def setup_heartbeats(interval)
             # suspended (e.g. after op7)
             if (@session && !@session.suspended?) || !@session
               sleep @heartbeat_interval
+              @bot.raise_heartbeat_event(Discordrb::Events::HeartbeatEvent.new(self))
```
meew0: This won't work as `self` should be a `Bot`, not a `Gateway`. Why not simply make `raise_heartbeat_event` create and raise the event, and just call `raise_heartbeat_event` with no argument here?

Daniel-Worrall: Yeah, I'm a dummy. Copying code without knowing what it does should be avoided :) In testing, it certainly raised, just didn't test the content. Oops.

Issue #225, opened 2016-09-16
Debug mode shows token in the log


Daniel-Worrall:
Security concern when people are absentmindedly sending logs to others, or if broadcasting their log for whatever reason

PR #231, opened 2016-09-22
Redacted the token from logging.

Daniel-Worrall:
Should fix all cases of #225


meew0:
This only prevents the token from being logged in fields called "token" - it would be more complete protection and simpler to just `gsub` the token in every packet that is printed.

Note that the analytics token is not the regular token - the worst an attacker can do with that is impersonate you in track calls. I'd prefer if that wasn't mentioned in discordrb in any way at all since analytics stuff is even farther up the "things libs shouldn't handle at all" list than, say, message acking or 2FA handling.

Also, this should be configurable as an initializer parameter, as somebody may *want* to have their token printed, for example for a complete log of packets sent they they may want to replay.

PR #233, opened 2016-09-27
Added user status support

Daniel-Worrall:


meew0:
Could you change the method names and comments to make it more clear that it's not the actual status being set, but rather the user setting? Right now it's pretty misleading in my opinion (especially as bots can't use these endpoints - there doesn't need to be a check for that but it should at least be mentioned in the comments)

I will also postpone this to 3.1.0 as the settings aren't really a must-have feature.

SnazzyPine25:
What about finding if the user is dnd? (`User#dnd?`) it seems to not be listed.

meew0:
That's not what this PR is about

PR #237, opened 2016-10-02
add user api to required files in bot.rb

shawila:
Not sure if it was left out on purpose, in which case this PR is useless :)


meew0:
Well, none of the `API::User` functions are actually used in `Bot`, so adding the require for that is pointless.

PR #243, opened 2016-10-08
Fixed logger token redacting on debug mode

Daniel-Worrall:
Fixes:
Cases of `#gsub!` on a Hash causing it to fail instead of log.
Cases of substituting the outgoing message token causing the authentication to break.


meew0:
If I'm reading this correctly, it will never log anything if a token *isn't* set, because the failing `if` condition will set `log` to `nil`.

PR #249, opened 2016-10-12
Add 'channels' attribute to commands and command_bot

Xzanth:
Adds an attribute 'channels' to commands that is for specifying the only channels that a command can be run on. Specified as any mix of channel names, id's and channel objects themselves.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -174,6 +174,7 @@ module Discordrb::Commands
       debug("Executing command #{name} with arguments #{arguments}")
       return unless @commands
       command = @commands[name]
+      return unless channels?(event.channel, command.attributes[:channels])
       unless command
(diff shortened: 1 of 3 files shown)


Xzanth:
Added a channels attribute to command bot as well, so one can contain a command bot to only respond to commands on a specific channel unless specified otherwise in the command's channel attribute.
Not sure if it fits to have this attribute, just wrote it how I'd ideally want to use it, happy to change.

z64:
Not sure if you saw the convo in `#ruby_discordrb`, would you be interested in also implementing the other similar "filters" for `CommandContainer`? 

i.e., `servers`, `users` ?

I'm not sure adding the bot initializer parameter fits with the design of the lib, but we'll see what meew has to say. If we stick with that, the existing `required_roles` should also be included in the initializer.

meew0:
As I said on Discord, I will merge this but please don't make any other filters like this unless absolutely necessary; eventually I will make the commands system more modular and then we can have all the filters we want.

PR #251, opened 2016-10-16
Allow the current game state to be cleared

ivgiuliani:
Context
----------

According to
https://discordapp.com/developers/docs/topics/gateway#gateway-status-update
it's possible to send a `null` value in place of a `game` object when
attempting to update the game's state. However, that has the effect of
*not* updating the current state. In other words, sending `null` will
keep the current state unchanged.

Problem
----------

It's currently impossible to reset the current game state. Steps to
reproduce:

1. change the game state via `bot.game="new game"`
2. the state is changed on discord
3. clear the state out via `bot.game = nil`
4. expected: the game state clears -- what happens instead: nothing

The reason is that we want to explicitely set `game.name = null` rather
than just `game = null` in the json we send to the server.

The fix
----------

I thought about using different approaches such as forcing to send the
payload if either `game` or `url` is different from their current state
set in the instance variables (respectively `@game` and `@streamurl`).
However that doesn't work if you want to reset the state at startup
since we have no knowledge of what those attributes were before.
The current fix always sends the payload upstream. It's not that many
bytes anyway.

Also I'm a bit confused by the `url` attribute. I can't find it anywhere
on the discord docs. Perhaps it's a legacy thing? Anyway I'm leaving it
unchanged for now.

Please let me know if this makes sense.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -465,7 +465,7 @@ module Discordrb
       @streamurl = url
       type = url ? 1 : nil
 
-      game_obj = game || url ? { name: game, url: url, type: type } : nil
+      game_obj = { name: game, url: url, type: type }
       @gateway.send_status_update(status, since, game_obj, afk)
     end
 
--- a/spec/bot_spec.rb
+++ b/spec/bot_spec.rb
@@ -1,0 +1,196 @@
+require 'discordrb'
+
+RSpec.describe Discordrb::Bot do
+  let(:gateway) { instance_double(Discordrb::Gateway) }
+  let(:token) { '1234567abcdef' }
+  let(:instance) { described_class.new(token: token) }
+
+  before do
+    instance.instance_variable_set(:@gateway, gateway)
+    allow(gateway).to receive(:open?).and_return(true)
+  end
+
+  describe 'when changing the current state' do
+    describe 'via #update_status' do
+      context 'when changing only the game' do
+        it 'attempts to send the new game' do
+          expect(gateway).to receive(:send_status_update)
+            .with(:online,
+                  anything,
+                  { name: 'new game', url: nil, type: nil },
+                  anything)
+
+          instance.update_status(:online, 'new game', nil)
+        end
+      end
+
+      context 'when changing only the stream url' do
+        it 'attempts to send the new stream url' do
+          expect(gateway).to receive(:send_status_update)
+            .with(:online,
+                  anything,
+                  { name: nil, url: 'new stream url', type: 1 },
+                  anything)
+
+          instance.update_status(:online, nil, 'new stream url')
+        end
+      end
+
+      context 'when changing both the game and the stream url' do
+        it 'attempts to send the new game object' do
+          expect(gateway).to receive(:send_status_update)
+            .with(:online,
+                  anything,
+                  { name: 'gamegame', url: 'new stream url', type: 1 },
+                  anything)
+
+          instance.update_status(:online, 'gamegame', 'new stream url')
+        end
+      end
+    end
+
+    describe 'via #game=' do
+      it 'attempts to send the new game' do
+        expect(gateway).to receive(:send_status_update)
+          .with(:online,
+                anything,
+                { name: 'new game', url: nil, type: nil },
+                anything)
+
+        instance.game = 'new game'
+      end
+
+      context 'when clearing out the currently set game' do
+        it 'sends a nil value upstream' do
+          expect(gateway).to receive(:send_status_update)
+            .with(:online,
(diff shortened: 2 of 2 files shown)


meew0:
First of all, this PR is one of the best-done PRs I've seen in a long time, good job on that! However, I can't reproduce your initial problem - running `bot.game = nil` works just fine on my side. This is a minimal example that works for me:

```rb
require "discordrb"

bot = Discordrb::Bot.new token: '<token>', client_id: <clientid>

bot.message(content: 'game test') do |event|
  bot.game = 'testing'
  sleep 1
  bot.game = nil
end

bot.run
```

Could you maybe show what code you were using for testing?

ivgiuliani:
I've got a repro on my project (https://github.com/ivgiuliani/discord-tfl/pull/39/files, possibly hard-ish to run locally but the key change is in `init.rb`). The only difference I can see is that `bot.game=` is called within a `ready` event. Other than that, the raw payload I send is identical to the one in your example (which however does work). Perhaps there's some difference in how `ready` events are processed; for completeness, the username change works all the time.

ivgiuliani:
I've got it down to a small repro:

```ruby
require 'discordrb'
CLEAR = !ARGV[0].nil? && ARGV[0] == "true"

bot = Discordrb::Bot.new token: TOKEN, client_id: CLIENT_ID
bot.ready do |event|
  if CLEAR
    puts "set nil"
    event.bot.game = nil
  else
    puts "set game"
    event.bot.game = "game"
  end
end

bot.run
```

(if run with `true` as first argument via command line it will attempt to clear the state, otherwise will set it to "game")

For me, running with `CLEAR=true` never unsets the previous game state. Perhaps there's  some race condition happening somewhere or possibly it's discord's fault. Will keep investigating.

Reproduction steps:
1. run `./test.rb`
2. game is changed
3. kill the script
4. run `./test.rb true`
5. nothing happens :/

meew0:
Ah, your problem is that you're trying to unset a game set on a different gateway connection. You can't do that, only overwrite it with a new status on the current connection; when that new status is unset it will revert again to the one set on the other connection. Consider these two bots:

```rb
bot = Discordrb::Bot.new # ...

bot.ready do
  bot.game = 'foo'
end

bot.run
```

```rb
bot = Discordrb::Bot.new # ...

bot.ready do
  bot.game = 'bar'
  sleep 1
  bot.game = nil
end

bot.run
```

Now you would run the first bot, wait until it sets the game, then run the second one. What you would expect is that the game switches from "foo" to "bar" to no game, however in reality it will switch from "foo" to "bar" to "foo".

If you kill the script, you cause the gateway to simply disconnect without sending close frames or the like, so Discord doesn't know that the script is gone before the next heartbeat times out. In the meantime, it assumes the connection is still alive, so it keeps displaying the bot user as online and with that game playing on Discord; the `game = nil` on the other script won't do anything regardless since no game to reset has been set on that gateway connection.

Now, the change this PR makes would make that work, but not because it actually unsets anything but because it overwrites the existing status again: the difference between `game: nil` and `game: {name: nil}` in the op3 packet is that the former unsets an existing presence from the current gateway connection, while the latter sets a new status without a name. The Discord client treats these two scenarios identically, but for the API, they aren't; in your example, if you kept the first bot running, then started up the second with the changes from this PR (so it would send an op3 with `name: nil` instead of `game: nil` for `bot.game = nil`), the game would unset on the client; however after either killing the second bot and waiting some time for Discord to notice the disconnection, or stopping the second bot cleanly using `bot.stop` or the likes, you would notice that the unset status reverts to the one originally set from the first bot. (Note that this scenario is somewhat hypothetical as discordrb doesn't currently handle received op1's correctly, so running two bots at once on the same account is always a little unstable, but if it did this is what would happen.)

Daniel-Worrall:
This is actually what I have been getting earlier and with more information I am able to reproduce it.

The exact scenarios where setting a game to nil does not work are when a game has not already been set during that session.

If the bot connects and there is already a game, then setting `bot.game = nil` will not clear it no matter how many times you try it. However, if you set the game (to either the same as it already is, or something else) with `bot.game = 'test'` for example then you are able to reset it back to `nil` with `bot.game = nil`

ivgiuliani:
That makes sense (sort of). Not saying this is the lib's fault, it's just a consequence of how discord works, but it's terribly confusing and counter intuitive. That said, thanks to your great explanation (thanks for that!) I managed to find a workaround (https://github.com/ivgiuliani/discord-tfl/pull/39/commits/5ef25fbdff6a14f5e8bb1cb2c96547d12b315046) which makes the issue less... visible. I'll close this PR as it's not really fixing anything.

PR #257, opened 2016-10-21
Add ignoring bot accounts as Bot param

greenbigfrog:
Not sure if we should also add it as command param. (Also not sure, if this is how you would like it implemented, probably not...)

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -91,11 +91,12 @@ module Discordrb
     # @param num_shards [Integer] The total number of shards that should be running. See
     #   https://github.com/hammerandchisel/discord-api-docs/issues/17 for how to do sharding.
     # @param redact_token [true, false] Whether the bot should redact the token in logs. Default is true.
+    # @param ignore_bots [true, false] Whether the bot should ignore Bot accounts or not. Default is false.
     def initialize(
         log_mode: :normal,
         token: nil, client_id: nil, application_id: nil,
         type: nil, name: '', fancy_log: false, suppress_ready: false, parse_self: false,
-        shard_id: nil, num_shards: nil, redact_token: true
+        shard_id: nil, num_shards: nil, redact_token: true, ignore_bots: false
     )
 
       LOGGER.mode = if log_mode.is_a? TrueClass # Specifically check for `true` because people might not have updated yet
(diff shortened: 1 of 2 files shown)


meew0:
Still needs to be added to the CommandBot initializer so it can be used there.

lib/discordrb/bot.rb
```diff
@@ -91,11 +91,12 @@ class Bot
     # @param num_shards [Integer] The total number of shards that should be running. See
     #   https://github.com/hammerandchisel/discord-api-docs/issues/17 for how to do sharding.
     # @param redact_token [true, false] Whether the bot should redact the token in logs. Default is true.
+    # @param ignore_bots [true, false] Whether the bot should ignore Bot accounts or not. Default is false.
```
meew0: "ignore bot accounts"

lib/discordrb/commands/command_bot.rb
```diff
@@ -66,6 +66,7 @@ class CommandBot < Discordrb::Bot
     #   :advanced_functionality). Default is '"'.
     # @option attributes [String] :quote_end Character that should end a quoted string (see
     #   :advanced_functionality). Default is '"'.
+    # @option attributes [true, false] :ignore_bots Whether the bot should ignore Bot accounts or not. Default is false.
```
meew0: "ignore bot accounts"

lib/discordrb/bot.rb
```diff
@@ -917,6 +919,11 @@ def handle_dispatch(type, data)
           return
         end
 
+        if @ignore_bots && user(data['author']['id']).bot_account?
```
meew0: Why does there need to be a resolution here? The author's data should already contain whether it is a bot or not.

lib/discordrb/bot.rb
```diff
@@ -917,6 +919,11 @@ def handle_dispatch(type, data)
           return
         end
 
+        if @ignore_bots && user(data['author']['bot'])
```
meew0: Did you even test this? `bot` isn't an ID.

greenbigfrog: lol...
yes, I did "test" it... But I know what went wrong when testing... was with my mind in crystal/js and fergot to build and install the gem
my bad

PR #258, opened 2016-10-23
fix indentaion of comment, that only own msgs can be edited

greenbigfrog:


meew0:
The description in a Yard comment doesn't need to be indented, only things like `@param` and `@return` do.

greenbigfrog:
stupid rubydoc

PR #266, opened 2016-11-13
Add REST support for reactions and abstractions for Message

z64:
These docs aren't live yet but can be viewed [here](https://github.com/hammerandchisel/discord-api-docs/blob/master/docs/resources/Channel.md#create-reaction--put-channelschanneliddocs_channelchannel-objectsmessagesmessageiddocs_channelmessage-objectreactionsemojidocs_channelemoji-structureme).

I added abstractions to `Message` that covers each of these endpoints. Not sure if there's anything else that would be more helpful.

I added `#to_reaction` to `Emoji` and `GlobalEmoji` in favor of interpolating the layout inside of the method. Hope that's okay, `#to_s` obviously can't facilitate this because of the leading /trailing `<:` `>`

### DIFF
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -150,6 +150,17 @@ module Discordrb::API::Channel
     )
   end
 
+  # Deletes all reactions on a message from all clients
+  def delete_all_reactions(token, channel_id, message_id)
+    Discordrb::API.request(
+      :channels_cid_messages_mid,
+      channel_id,
+      :delete,
+      "#{Discordrb::API.api_base}/channels/#{channel_id}/messages/#{message_id}/reactions",
+      Authorization: token
+    )
+  end
+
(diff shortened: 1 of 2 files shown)


lib/discordrb/api/channel.rb
```diff
@@ -133,6 +133,70 @@ def bulk_delete_messages(token, channel_id, messages = [])
     )
   end
 
+  # TODO: Insert docs links once they are pushed to master
+  # for reactions endpoints
```
greenbigfrog: Maybe link it already. Or just remove it for now... or not have the PR merged until it's fully public/pushed/whatever!?

z64: Yeah. I wasn't about to try and figure out what the URLs should be, or have them be wrong and have to fix them later (or worse, someone forgets to), so I made the note. I'm not sure when the docs on the site are updated from master, either

PR #270, opened 2016-11-14
Typing events can be responded to like any other message

VxJasonxV:
This change was spurred by 'drewajp':

screen shot 2016-11-14 at 12 56 07 am

A reference code example:

```ruby
require 'discordrb'

bot = Discordrb::Bot.new token: ENV['DISCORD_TOKEN'], client_id: ENV['DISCORD_APPLICATION_ID']

bot.typing do |event|
  if event.user.id == ENV['DISCORD_ADMIN_USER_ID'].to_i
    event.respond %Q{You started typing at #{event.timestamp}}
  end
end

bot.run
```

When running this, the bot sent

```plain
You started typing at 2016-11-14 07:49:59 +0000
```

In the channel I started typing in. Tested it on two different Discord servers, one channel named `#general`, one not.

### DIFF
--- a/lib/discordrb/events/typing.rb
+++ b/lib/discordrb/events/typing.rb
@@ -5,6 +5,7 @@ require 'discordrb/events/generic'
 module Discordrb::Events
   # Event raised when a user starts typing
   class TypingEvent < Event
+    include Respondable
     # @return [Channel] the channel on which a user started typing.
     attr_reader :channel
 
--- a/lib/discordrb/events/typing.rb
+++ b/lib/discordrb/events/typing.rb
@@ -6,6 +6,7 @@ module Discordrb::Events
   # Event raised when a user starts typing
   class TypingEvent < Event
     include Respondable
+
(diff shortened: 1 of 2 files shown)


beezTEM:
We need this because how many times do you think this might be helpful?

lib/discordrb/events/typing.rb
```diff
@@ -5,6 +5,7 @@
 module Discordrb::Events
   # Event raised when a user starts typing
   class TypingEvent < Event
+    include Respondable
```
meew0: Please add a newline after this; while RuboCop doesn't require it I think it's good style to separate the included modules like that.

VxJasonxV: screen shot 2016-11-14 at 04 35 01

;D

VxJasonxV:
Fixed not only my addition, but the two other locations where it was missing a newline too.

meew0:
I didn't even know it was like that everywhere, thanks for finding that out!

PR #275, opened 2016-11-17
Limits commands in help list

FormalHellhound:
This limits the commands in the help list to only show commands a user has the correct role or permissions to access. This was done by https://github.com/z64/ (Lune), but he had no intention of submitting a pull request, and told me to mention him if I submitted a pull for it. It does not support permission_level.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -152,7 +152,7 @@ module Discordrb::Commands
           result
         else
           available_commands = @commands.values.reject do |c|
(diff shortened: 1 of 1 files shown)


lib/discordrb/commands/command_bot.rb
```diff
@@ -151,7 +151,9 @@ def initialize(attributes = {})
           end
           result
         else
-          available_commands = @commands.values.reject { |c| !c.attributes[:help_available] }
+          available_commands = @commands.values.reject do |c|
+            !c.attributes[:help_available] || !required_roles?(event.author, c.attributes[:required_roles]) || !required_permissions?(event.author, c.attributes[:required_permissions], event.channel)
```
z64: Could you use `event.user` instead of `event.author` to be consistent with the rest of the block?

PR #278, opened 2016-12-04
Use bundler, not gem install

VxJasonxV:
The use of `gem install` is contains potential difficulty depending on the state of the user's existing development environment, not to mention the difficulty of `gem install` vs `sudo gem install`, permissions errors, and plenty of other common gotchas. Bundler installs gems into userspace and integrates nicely into managed ruby environments (rbenv/rvm).

The use of Bundler is highly recommended over the standard `gem install` process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ Using [Bundler](http://bundler.io/#getting-started), you can add discordrb to yo
 
     gem 'discordrb'
 
+And then install via `bundle install`.
+
 ### Windows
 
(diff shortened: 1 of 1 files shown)


VxJasonxV:
I did some research into Bundler and Windows, but I couldn't find if Bundler has a setting analogous to `--platform=ruby`. I would greatly prefer if Bundler would be recommended for all environments.

meew0:
I'm fine with recommending bundler by default but I think the vanilla Ruby installation method should also be mentioned as an alternative. How about something like "Alternatively, while Bundler is the recommended option, you can also install discordrb without it:" and then the original stuff?

z64:
Continued in #386

PR #282, opened 2016-12-05
fix/update return type for `Server#emoji`

### LINKED ISSUE #281: Method emoji of Server is not correctly documented
The documentation generated on master says that `server.emoji` return an `Array `of `Emojis`.

image

Or this piece of code : 
`` event.server.emoji.each { |e| puts "#{e} #{e.class}" }``

will say that `e` is an `Array` and not an `Emoji` : 
image

I don't know if this behavior is intended, or if it is an error in the documentation.

greenbigfrog:
closes #281

### DIFF
--- a/lib/discordrb/data.rb
+++ b/lib/discordrb/data.rb
@@ -2141,7 +2141,7 @@ module Discordrb
     # @return [Array<Role>] an array of all the roles created on this server.
     attr_reader :roles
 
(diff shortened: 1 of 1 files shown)


Daniel-Worrall:
This has only changed the object return information. It still states that the return is an array in the description.

greenbigfrog:
nice catch nekka

PR #283, opened 2016-12-05
fix Bot#emoji(ID)

greenbigfrog:


lib/discordrb/bot.rb
```diff
@@ -160,15 +160,16 @@ def servers
...
       gateway_check
       if id
-        emoji = @emoji.find { |sth| sth.id == id }
+        emoji
+        @emoji.find { |sth| sth.id == id }
       else
         emoji = {}
         @servers.each do |_, server|
           server.emoji.values.each do |element|
             emoji[element.name] = GlobalEmoji.new(element, self)
```
z64: Are these really `GlobalEmoji` though? (`Emoji that is not tailored to a server`)
I could understand if `Emoji` inherited `GlobalEmoji` - in fact, they are virtually the same, I think how it's implemented currently is redundant design - but how are you making the distinction that the emoji that you're storing here are, in fact, `Global` and not just "regular" `Emoji`?

PR #286, opened 2016-12-07
Similar code refactoring

Daniel-Worrall:
A first attempt at refactoring very similar code to keep things DRY.

Also includes a minor change to a log where the order of the user and server id was wrong

E: Adding Rspec check as this will need proper testing to ensure everything still works as intended.

- [ ] RSpec
- [ ] Safe to Merge

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -122,6 +122,12 @@ module Discordrb::API
     response
   end
 
+  # Generic API request helper
+  def generic_request(token, id, path, key, type, data = nil)
+    return request(key, id, type, "#{api_base}/#{path}", Authorization: token) if data.nil?
+    request(key, id, type, "#{api_base}/#{path}", data, Authorization: token, content_type: :json)
+  end
+
   # Make an icon URL from server and icon IDs
   def icon_url(server_id, icon_id)
     "#{api_base}/guilds/#{server_id}/icons/#{icon_id}.jpg"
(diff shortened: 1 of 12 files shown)


meew0:
Thanks for going through all this effort! Tell me once you consider this finished, I'll take a close look at it then. In the meantime I'll label it as long-term, take your time

lib/discordrb/api.rb
```diff
@@ -122,6 +122,12 @@ def request(key, major_parameter, type, *attributes)
     response
   end
 
+  # Generic API request helper
+  def generic_request(token, id, path, key, type, data = nil)
+    return request(key, id, type, "#{api_base}/#{path}", Authorization: token) if data.nil?
+    request(key, id, type, "#{api_base}/#{path}", data, Authorization: token, content_type: :json)
```
LikeLakers2: Since this is a PR all about removing similar code, I'm going to recommend that the `.to_json` on all those hashes in all those other `API` methods be moved to this method. :)

z64:
Including existing conflicts, #390 will make changes to `data.rb` conflicting as well.

Would you like to try to get this merged first before we split `data.rb`, or revisit it afterwards? @Daniel-Worrall

Daniel-Worrall:
Yup, this is definitely on hold for now. Things like the data.rb rewrite should come first, then I'll revisit this. A lot of the things I did in this PR should be thrown out and changed for better code too as I didn't know what I was doing at the time.

I'll work on figuring out what can/should be refactored in the coming weeks.

z64:
Closed per discussion

PR #298, opened 2017-01-06
Command argument type checking

fwrs:
This PR allows to define types for command arguments which improves the experience of implementing new commands. For example, a basic multiplication command
```rb
bot.command(:multiply) do |event, a, b|
  begin
    Integer(a) * Integer(b)
  rescue ArgumentError => _
    'Arguments should be numbers'
  end
end
```
with this update looks like this:
```rb
bot.command(:multiply, arg_types: [Integer, Integer]) do |e, a, b|
  break 'Arguments should be numbers' unless a && b
  a * b
end
```


lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,54 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
...
+        when types[i] == Discordrb::Role
+          result = parse_mention arg, server
+          result if result.instance_of? Discordrb::Role
+        when types[i] == Discordrb::Emoji
+          result = parse_mention arg, server
+          result if result.instance_of? Discordrb::Emoji
+        when types[i] == Discordrb::Invite
+          resolve_invite_code arg
+        when types[i].respond_to?(:from_argument)
+          types[i].from_argument arg rescue nil
```
meew0: There should probably be an `else` case here that raises an error if an invalid type is defined.

lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,54 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
...
+    def arg_check(args, types = nil, server = nil)
+      return args unless types
+      args.each_with_index.map do |arg, i|
+        break arg unless types[i] || types[i] == String
+        case
+        when types[i] == Integer
+          Integer(arg) rescue nil
+        when types[i] == Float
+          Float(arg) rescue nil
+        when types[i] == Timeend
```
meew0: `Timeend`?

lib/discordrb/commands/command_bot.rb
```diff
@@ -190,6 +191,7 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
         return
       end
       return unless !check_permissions || channels?(event.channel, command.attributes[:channels])
+      arguments = arg_check(arguments, command.attributes[:args])
```
meew0: In my opinion the name `args` is bad (too general), how about `types` or `arg_types`?

lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,54 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
       raise
     end
 
+    # Transforms an array of string arguments based on types array.
+    # For example, `['1', '10..14']` with types `[Integer, Range]` would turn into `[1, 10..14]`.
+    def arg_check(args, types = nil, server = nil)
+      return args unless types
+      args.each_with_index.map do |arg, i|
+        break arg unless types[i] || types[i] == String
```
meew0: This should be `next` not `break`

meew0: Also this logic seems redundant, it will ignore the argument unless it (exists or is `String`). If the argument is `String`, it will exist.

If you meant something like `unless it exists, or if it is a string`, you might want to express that. I'd do `if types[i].nil? || types[i] == String`, but if you find something more concise, feel free to use it

lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,54 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
       raise
     end
 
+    # Transforms an array of string arguments based on types array.
+    # For example, `['1', '10..14']` with types `[Integer, Range]` would turn into `[1, 10..14]`.
+    def arg_check(args, types = nil, server = nil)
```
meew0: This method is never called with a `server`, so this argument is unnecessary (but you should probably change the invocation instead to use it)

lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,54 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
...
+        when types[i] == Symbol
+          arg.to_sym
+        when types[i] == Encoding
+          Encoding.find arg rescue nil
+        when types[i] == Regexp
+          Regexp.new arg rescue nil
+        when types[i] == Rational
+          Rational(arg) rescue nil
+        when types[i] == Range
+          Range.new(*arg.split("..").map(&:to_i)) rescue nil
```
meew0: This doesn't account for exclusive ranges (`1...4`)

lib/discordrb/commands/command_bot.rb
```diff
@@ -207,6 +209,85 @@ def execute_command(name, event, arguments, chained = false, check_permissions =
...
+        elsif types[i] == Rational
+          begin
+            Rational(arg)
+          rescue ArgumentError
+            nil
+          end
+        elsif types[i] == Range
+          begin
+            if arg.include? '...'
+              Range.new(*arg.split('...').map(&:to_i))
```
meew0: Exclusive ranges still aren't handled correctly; this case should set the `exclude_end` parameter to true.

PR #303, opened 2017-01-12
Fix filtering of playing events by game or type

sven-strothoff:
There are two small issues in the PlayingEventHandler that result in crashes if a filter for game or type is used.

During the initialisation of PlayingEvent the game _object_ received from the server is flattened to its only attribute `name` (presence.rb:75). In line 100 `event.game.name` is accessed, but at this point `event.game` is only a string.

For filtering by event type the `type` attribute is accessed twice (i.e. `event.type.type`) in lines 105 and 107.

### DIFF
--- a/lib/discordrb/events/presence.rb
+++ b/lib/discordrb/events/presence.rb
@@ -96,18 +96,10 @@ module Discordrb::Events
                end
         end,
         matches_all(@attributes[:game], event.game) do |a, e|
-          a == if a.is_a? String
-                 e.name
-               else
-                 e
-               end
+          a == e
         end,
         matches_all(@attributes[:type], event.type) do |a, e|
-          a == if a.is_a? Integer
-                 e.type
-               else
-                 e
-               end
+          a == e
         end
       ].reduce(true, &:&)
     end


lib/discordrb/events/presence.rb
```diff
@@ -97,14 +97,14 @@ def matches?(event)
         end,
         matches_all(@attributes[:game], event.game) do |a, e|
           a == if a.is_a? String
-                 e.name
+                 e
                else
                  e
                end
```
Daniel-Worrall: This call all just be written as `a == e`

lib/discordrb/events/presence.rb
```diff
@@ -97,14 +97,14 @@ def matches?(event)
...
         end,
         matches_all(@attributes[:type], event.type) do |a, e|
           a == if a.is_a? Integer
-                 e.type
+                 e
                else
                  e
                end
```
Daniel-Worrall: and here

meew0:
Nice catch!

sven-strothoff:
Yeah, I had a strange feeling about the `e` in the if and else branches. I thought it mattered that a is a String/Integer respectively. But you are right @Daniel-Worrall, it's just `e` in every case. I amended my commit. Thank you!

Issue #309, opened 2017-01-30
Missing API methods, as well as some missing parameters


LikeLakers2:
Hi there, whilst browsing discordapp.com's API section, I happened to notice some methods that had parameters that could be specified, that do not have equivalent parameters in discordrb's API modules. So I went through, and *hoo boy* have I found quite a lot. I even found some API endpoints that do not have an equivalent API module method. So have a list.

Where possible, I do not take into account the reason (if any may exist) that a method may not have a parameter. I simply looked if the official API documentation and discordrb had the same parameters available for use.

### Missing parameters
- [x] `API::Channel.messages`, missing `around_id` parameter -- #314
- [x] `API::Channel.create_message`, missing `nonce` parameter -- #414 
- [x] `API::Channel.create_invite`, missing `unique` parameter -- #312
- [x] `API::Server.update`, missing `verification_level`, `default_message_notifications`, ~~`owner_id`~~, `splash` parameters -- #414 
- [x] `API::Server.create_channel`, missing `bitrate`, `user_limit`, `permission_overwrites` parameters -- #321 
- [x] `API::Server.create_role`, missing `name`, `permissions`, `color`, `hoist`, `mentionable` parameters -- #323 
- [x] `API::Server.prune_count`, missing `days` parameter -- #315

### Missing API module equivalents
As mentioned, I have found several API endpoints documented on Discord's API documentation that do not have an equivalent in discordrb's API modules.

Guild:
- [x] [Add Guild Member](https://discordapp.com/developers/docs/resources/guild#add-guild-member) --
 #413 
- [x] [Add Guild Member Role](https://discordapp.com/developers/docs/resources/guild#add-guild-member-role)/[Remove Guild Member Role](https://discordapp.com/developers/docs/resources/guild#remove-guild-member-role) -- #310
- [x] [Modify Guild Role Positions](https://discordapp.com/developers/docs/resources/guild#modify-guild-role-positions) -- #445
- [x] [Get Guild Voice Regions](https://discordapp.com/developers/docs/resources/guild#get-guild-voice-regions) -- #311 
- [x] [Get Guild Embed](https://discordapp.com/developers/docs/resources/guild#get-guild-embed)/[Modify Guild Embed](https://discordapp.com/developers/docs/resources/guild#modify-guild-embed) -- #435 
- [x] Abstraction for `API::Server.invites` -- #394 
- [x] `ServerBan` or `Ban` class that supports `reason` and `#remove` for lifting bans -- #404 

Channels:
[...]

PR #310, opened 2017-01-31
Support Add/Remove Guild Member Role

z64:
WIP ; Need to test (and rebase the last 'oops' commits into 542ac2b / 8b0e94d)

Opted to not increase the complexity of the other role modification methods.

#309

### DIFF
--- a/lib/discordrb/api/server.rb
+++ b/lib/discordrb/api/server.rb
@@ -254,6 +254,19 @@ module Discordrb::API::Server
     )
   end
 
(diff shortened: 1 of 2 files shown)


z64:
Annnd tested both endpoints :+1: 

Should be OK to merge if no other implementations of this endpoint are wanted elsewhere that I might have missed.

(If I try to appease codeclimate here, it would be conflicting with Nekka's WIP PR, so I'll keep it the same and leave that to him)

PR #313, opened 2017-01-31
Small resolve_id cleanup

LikeLakers2:
Just noticed a couple methods still using `.id` where they could easily use `.resolve_id`.

Of note, I also found [Channel#define_overwrite](https://github.com/LikeLakers2/discordrb/blob/master/lib/discordrb/data.rb#L1265-L1294) to be using `.id` where it *potentially* could be using `.resolve_id`, however I am unsure how to deal with this due to the class check within that method making things awkward, at least without doing some searches through the `@users` or `@roles` arrays, which I'm unsure if we want.


meew0:
Squashed and merged because of the "Merge remote-tracking branch" commits

PR #314, opened 2017-02-01
Around ID in channel history

z64:
#309


meew0:
The `around_id` parameter should be documented in `Channel#history`.

PR #322, opened 2017-02-12
Make Permissions Great Again

z64:
~~*(Sorry/not sorry for the tired meme title)*~~ **I'm sorry.**

**Dealing with `Permissions` sucks.** This gets rid of the "boilerplate" of flipping bits and makes it suck less.

## Task: Shipping a new channel with overwrites
(This example depends on #321)
### Before

```ruby
allow = Discordrb::Permissions.new
allow.can_mention_everyone = true
allow.can_send_tts_messages = true

deny = Discordrb::Permissions.new
deny.can_create_instant_invite = true

# (let's just say we already have a Server and a Role)
overwrite = Discordrb::Overwrite.new(role, allow: allow, deny: deny)
server.create_channel('meme channel', 0, permission_overwrites: [overwrite])
```

### After

```ruby
overwrite = Discordrb::Overwrite.new role, allow: [:mention_everyone, :send_tts], deny: [:create_instant_invite]
server.create_channel('meme channel', 0, permission_overwrites: [overwrite])
```

## Task: Interacting with the REST API

Previously you would have to do something similar (create a `Permissions` instance) just to get some bits to ship to the API. This PR exposes a helper if you just want the bits and makes the API easier to use directly.

### After

```ruby
Discordrb::API::Channel.edit_channel_permissions(
  bot.token, 
  225375815087554563, 
  120571255635181568, 
  Discordrb::Permissions.bits([:read_messages, :attach_files]), 
  Discordrb::Permissions.bits([:mention_everyone, :use_external_emoji]), 
  'member'
)
```

This of course has implications in other places in the library where permissions are required. Methods could take an array of symbols and use the new initializer case to make `Permissions`, or users can use `Permissions.bits` for anything that doesn't but will take bits anyway. 

---

Todo:

- [x] Specs! :heart: 
- [x] Add `:administrate` alias

### DIFF
--- a/spec/permissions_spec.rb
+++ b/spec/permissions_spec.rb
@@ -1,0 +1,76 @@
+require 'discordrb'
+
+module Discordrb
+  describe Permissions do
+    subject { Permissions.new }
+
+    describe Permissions::Flags do
+      it 'creates a setter for each flag' do
+        responds_to_methods = Permissions::Flags.map do |_, flag|
+          subject.respond_to?(:"can_#{flag}=")
+        end
+
+        expect(responds_to_methods.all?).to eq true
+      end
+
+      it 'calls #write on its writer' do
+        writer = double
+        expect(writer).to receive(:write)
+
+        Permissions.new(0, writer).can_read_messages = true
+      end
+    end
+
+    context 'with Flags stubbed' do
+      before do
+        stub_const('Discordrb::Permissions::Flags', 0 => :foo, 1 => :bar)
+      end
+
+      describe '#init_vars' do
+        it 'sets an attribute for each flag' do
+          expect(
+            [
+              subject.instance_variable_get('@foo'),
+              subject.instance_variable_get('@bar')
+            ]
+          ).to eq [false, false]
+        end
+      end
+
+      describe '.bits' do
+        it 'returns the correct packed bits from an array of symbols' do
+          expect(Permissions.bits(%i[foo bar])).to eq 3
+        end
+      end
+
+      describe '#bits=' do
+        it 'updates the cached value' do
+          allow(subject).to receive(:init_vars)
+          subject.bits = 1
+          expect(subject.bits).to eq(1)
+        end
+
+        it 'calls #init_vars' do
+          expect(subject).to receive(:init_vars)
+          subject.bits = 0
+        end
+      end
+
+      describe '#initialize' do
+        it 'initializes with 0 bits' do
+          expect(subject.bits).to eq 0
+        end
+
+        it 'can initialize with an array of symbols' do
+          instance = Permissions.new %i[foo bar]
+          expect(instance.bits).to eq 3
+        end
+
+        it 'calls #init_vars' do
+          expect_any_instance_of(Permissions).to receive(:init_vars)
+          subject
+        end
+      end
+    end
+  end
+end
--- a/lib/discordrb/permissions.rb
+++ b/lib/discordrb/permissions.rb
@@ -74,13 +74,43 @@ module Discordrb
       end
     end
 
+    # Return the corresponding bits for an array of permission flag symbols.
+    # This is a class method that can be used to calculate bits instead
+    # of instancing a new Permissions object.
+    # @example Get the bits for permissions that could allow/deny read messages, connect, and speak
+    #   Permissions.bits [:read_messages, :connect, :speak] #=> 3146752
+    # @param list [Array<Symbol>]
+    # @return [Integer] the computed permissions integer
+    def self.bits(list)
+      value = 0
+
+      Flags.each do |position, flag|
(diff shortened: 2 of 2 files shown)


meew0:
Would you mind adding some specs for this? It should be fairly easy to do given that this PR doesn't involve any API calls.

z64:
Sure, I'll try.

z64:
@Daniel-Worrall I've pushed the specs for this, if you could please take a look :smile: 

I was at another impasse about calling `#can_read_messages` for the `calls #write on its writer` spec, but I wasn't sure the best way about it. I didn't feel like doing another thing like `Flags.each` was appropriate here.

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        responds_to_methods = Permissions::Flags.map do |_, flag|
+          default_instance.respond_to?(:"can_#{flag}=")
+        end
+
+        expect(responds_to_methods.all?).to eq true
+      end
+
+      it 'calls #write on its writer' do
+        writer = double
+        allow(writer).to receive(:write)
```
Daniel-Worrall: This allow is redundant. The expect below does the exact same thing except it also adds an expectation that it should receive it and will fail if it doesn't.

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
+module Discordrb
```
Daniel-Worrall: Can you add a `require 'discordrb'` to be consistent and allow individual spec files to be tested at one time?

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        expect(responds_to_methods.all?).to eq true
+      end
+
+      it 'calls #write on its writer' do
+        writer = double
+        allow(writer).to receive(:write)
+        expect(writer).to receive(:write)
+
+        instance = Permissions.new(0, writer)
+        instance.can_read_messages = true
```
Daniel-Worrall: I feel like these should just be considered as one and not needed to be set to a variable.

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+          expect(default_instance.bits).to eq 0
+        end
+
+        it 'can initialize with an array of symbols' do
+          instance = Permissions.new %i[foo bar]
+          expect(instance.bits).to eq 3
+        end
+
+        it 'calls #init_vars' do
+          allow_any_instance_of(Permissions).to receive(:init_vars)
```
Daniel-Worrall: Another redundant allow

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      describe '#bits=' do
+        it 'updates the cached value' do
+          allow(default_instance).to receive(:init_vars)
+          default_instance.bits = 1
+          expect(default_instance.bits).to eq(1)
+        end
+
+        it 'calls #init_vars' do
+          allow(default_instance).to receive(:init_vars)
```
Daniel-Worrall: Another redundant allow

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+      describe '.bits' do
+        it 'returns the correct packed bits from an array of symbols' do
+          expect(Permissions.bits(%i[foo bar])).to eq 3
+        end
+      end
+
+      describe '#bits=' do
+        it 'updates the cached value' do
+          allow(default_instance).to receive(:init_vars)
```
Daniel-Worrall: This is not needed and actually adds processing time to the spec. Since init_vars is a really basic method that doesn't have any outside dependencies, it's heavier to stub it through the allow than it is to just let it run.

spec/permissions_spec.rb
```diff
@@ -0,0 +1,78 @@
+module Discordrb
+  describe Permissions do
+    subject(:default_instance) { Permissions.new }
```
Daniel-Worrall: I don't like giving the subject a name that isn't directly related to the name of the object. I would prefer this be called permissions or left blank and just referred to as subject.

spec/permissions_spec.rb
z64: Actually, now that I think about it (although I've addressed your comment in the fixup commit), isn't the convention to put this in the spec helper file and then all specs should require that? Maybe make `spec_helper.rb` require it in a future refactor?

Daniel-Worrall: I would do this, except the repo now houses the webhook gem and the main gem and are tested independently.

spec/permissions_spec.rb
Daniel-Worrall: Thinking about this later, technically this should remain as it introduces an outside dependency if left out but at this point I don't see issue with it not being there.

PR #325, opened 2017-02-18
Update MessageLength Error to reflect API change

Daniel-Worrall:
Using Postman, the current response on this kind of error gives
```
{
  "content": [
    "Must be 2000 or fewer characters long."
  ]
}
```

### DIFF
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -78,7 +78,7 @@ module Discordrb::API::Channel
     )
   rescue RestClient::BadRequest => e
     parsed = JSON.parse(e.response.body)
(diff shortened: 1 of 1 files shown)


LikeLakers2:
I do wonder why we rely on checking for this exact string, rather than, I don't know, maybe just seeing if `parsed['content'].is_a?(Array)` and leaving it at that.

z64:
Because it could have failed for another reason, other than the message being too long, which would not fit  `Discordrb::Errors::MessageTooLong` that is raised.

LikeLakers2:
Would there be any other way to detect this, then? I really don't believe relying on a static string to stay updated is a good method here.

PR #328, opened 2017-02-26
Some more resolve_id cleanup + a couple other things

LikeLakers2:
Just cleaning up some more stuff in DRB that I found. Nothing too much, besides maybe a small potential performance improvement in `Channel#users` that takes advantage of `Channel#default_channel?`.


LikeLakers2:
After noticing and correcting an inheritance error via b32e20d1a912f6c47abe50983fc3e28cc440ff89, I am considering cleaning up the `Discordrb::Events` namespace a little.

If I do so, I'll commit all that work all at once since it will be a little more invasive and potentially a little breaking if merged before it's finished -- so if you feel the need to merge this PR before I'm done, feel free and I'll just throw the commits into a new branch.

LikeLakers2:
Scratch the above comment, I will make the event stuff a separate PR to be handled later. It's become more than just a cleanup with what I feel I will want to do, so I feel it doesn't fit under this PR anymore.

PR #329, opened 2017-03-06
add Event inheritance for reaction events

valeth:
add for ReactionEvent and ReactionAddEvent


LikeLakers2:
I would like to mention that I already did this in https://github.com/meew0/discordrb/pull/328/commits/b32e20d1a912f6c47abe50983fc3e28cc440ff89. Though I doubt it really matters since I don't believe it'll cause a conflict.

PR #338, opened 2017-04-28
[WIP] Add optional error checking to Webhooks::Embeds

z64:
We get a lot of questions about Embeds because people are getting `400 Bad Request`. We could make a wiki page, add examples, but I think it should be easy enough to have the library tell them what they are doing wrong, and stop the request before it gets sent to Discord. (this extends to users of `discordrb-webhooks` too)

The limitations of embeds aren't documented, as I'm certain they could change at any time - so I've made this checking optional under a `check_format` named argument that defaults to `true`, so it can be turned off in case it ever becomes inaccurate / before discordrb contributor can update the limits.

Just wanted to open this PR to get any early feedback. This still has debug code and memes inside of it.

I need to:
- [ ] Test that each of these limits are accurate as documented by the community
- [ ] Figure out why the `require 'discordrb/webhooks/errors'` fails (??) and actually use my exception class
- [ ] See that there aren't any other additional limits I missed (comments welcome!)
- [ ] This primarily handles Embeds, but can extend to Webhooks too. That said, the only similar limitation I know of is that Webhook posts can have only up to 5 embeds. I've not implemented this exception yet.
- [ ] Remove memes
- [ ] Document `check_format`
- [ ] Pass rubocop, as I know its going to fail when I hit "Create pull request"

I've opted to just use one exception (`EmbedFormatError`) and build a string of all the errors found in the embed, and then only `raise` once. This is opposed to making a `raise` for each property, where you would have to keep retrying to lint more errors in the embed... this way you can hopefully fix them all in one go. Perhaps this isn't appropriate though? I'm not sure.

This also uses safe navigation.

Examples:

### DIFF
--- a/lib/discordrb/webhooks/embeds.rb
+++ b/lib/discordrb/webhooks/embeds.rb
@@ -1,9 +1,38 @@
+puts 'lol'
+
 module Discordrb::Webhooks
   # An embed is a multipart-style attachment to a webhook message that can have a variety of different purposes and
   # appearances.
   class Embed
+    # Maximum title length
+    MAXIMUM_TITLE_LENGTH = 256
+
+    # Maximum number of fields the embed can carry
+    MAXIMUM_FIELD_COUNT = 25
+
+    # Maximum length of the description field
+    MAXIMUM_DESCRIPTION_LENGTH = 2048
+
+    # Maximum footer length
+    MAXIMUM_FOOTER_LENGTH = 2048
+
+    # Maximum author length
+    MAXIMUM_AUTHOR_NAME_LENGTH = 256
+
+    # Maximum field name length
+    MAXIMUM_FIELD_NAME_LENGTH = 256
+
+    # Maximum field value length
+    MAXIMUM_FIELD_VALUE_LENGTH = 1024
+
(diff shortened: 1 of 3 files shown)


lib/discordrb/webhooks/embeds.rb
```diff
@@ -1,9 +1,38 @@
...
+    MAXIMUM_AUTHOR_NAME_LENGTH = 256
+
+    # Maximum field name length
+    MAXIMUM_FIELD_NAME_LENGTH = 256
+
+    # Maximum field value length
+    MAXIMUM_FIELD_VALUE_LENGTH = 1024
+
+    # Maximum total accounted length
+    MAXIMUM_LENGTH = 4000
```
soukouki: > In addition to the limits above, the sum of all characters in an embed structure must not exceed *6000* characters.

z64: This PR was made before that limitation was documented.

z64:
I'm not confident our story on request validation is clear enough yet to justify continuing this right now. I'd like to see v7 error parsing first.

PR #340, opened 2017-04-29
Fix Role#members

z64:


meew0:
How did this even happen

PR #345, opened 2017-05-18
Make embed a symbol in Channel#send_message

greenbigfrog:


greenbigfrog:
I shouldn't PR past midnight

LikeLakers2:
I was gonna say, what's wrong with this PR? Making it a symbol seems perfectly reasonable to me, I'm curious why you closed this.

z64:
This was done as a follow up to a conversation in the channel about making *tts* a named argument, or at least moving it to the end of the arg, because no one uses TTS and it just makes the method call ugly.

Gbf however thought we were talking about making `embed` named, hence his comment and closing.

This also breaks bots.

LikeLakers2:
Understood. I understand it can break bots, but I'd be in support of this even if it's kept on hold for a major update.

PR #350, opened 2017-05-29
Reaction Events Parsing on Self

Daniel-Worrall:
- [ ] Specs

This prevents the events from firing when the bot is set to not parse self.

Let me know if this should be added to any other events.


Daniel-Worrall:
I could work those into a single method, but I don't see the saved complexity.

PR #351, opened 2017-05-29
Extension on Reaction events to return Member

Daniel-Worrall:
- [ ] ~~Specs~~ Won't add yet

### DIFF
--- a/lib/discordrb/events/reactions.rb
+++ b/lib/discordrb/events/reactions.rb
@@ -24,7 +24,7 @@ module Discordrb::Events
     def user
       # Cache the user so we don't do requests all the time
       @user ||= if server
-                  @bot.user(@user_id).on(@server)
+                  @server.member(@user_id)
                 else
                   @bot.user(@user_id)
                 end
--- a/lib/discordrb/events/reactions.rb
+++ b/lib/discordrb/events/reactions.rb
@@ -23,13 +23,11 @@ module Discordrb::Events
     # @return [User, Member] the user that reacted to this message, or member if a server exists.
     def user
       # Cache the user so we don't do requests all the time
-      if @user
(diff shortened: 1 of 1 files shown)


lib/discordrb/events/reactions.rb
```diff
@@ -20,10 +20,16 @@ def initialize(data, bot)
...
     def user
       # Cache the user so we don't do requests all the time
-      @user ||= @bot.user(@user_id)
+      if @user
+        @user
+      elsif server
+        @user = @bot.user(@user_id).on(@server)
+      else
+        @user = @bot.user(@user_id)
+      end
```
z64: I believe this could be neatened to:
```
@user ||= if server
            server.member(@user_id)
          else
            @bot.user(@user_id)
          end
```

Daniel-Worrall:
This is good to merge. I'm abstaining on the specs for now due to thinking about a better method of providing spec fixtures than the structure we have right now. Tested with eval code

lib/discordrb/events/reactions.rb
```diff
@@ -20,10 +20,14 @@ def initialize(data, bot)
       @channel_id = data['channel_id'].to_i
     end
 
-    # @return [User] the user that reacted to this message.
+    # @return [User, Member] the user that reacted to this message, or member if a server exists.
     def user
       # Cache the user so we don't do requests all the time
-      @user ||= @bot.user(@user_id)
+      @user ||= if server
+                  @bot.user(@user_id).on(@server)
```
meew0: This can be simplified to `@server.member(@user_id)`

z64: I told him to do that but Nekka doesn't like me.

PR #358, opened 2017-06-01
Use CDN URL for avatars and server icons

z64:
This also allows us to pull GIF avatars. Yay!


z64:
Just a small update on this - apparently Discord's embeds will no longer embed resources that the current master is serving via the API routes. Only CDN URLs appear to embed properly now. Use cases that create something like a "user profile" embed are "broken" in this regard.

PR #360, opened 2017-06-06
Add :rescue command attribute for lazy exception handling

z64:
It is useful to perform follow up actions when unexpected errors occur internally. This adds a command attribute that will be `#called` whenever something unhandled by the command's `@block` is raised, or alternatively a String to respond to the `event` with.

Example:
```ruby
bot.command(:test, rescue: "oops! %exception%") { 1 / 0 }
```
image

With a custom proc:
```ruby
rescue_proc = proc do |event, exception|
  BOT.user(120571255635181568).pm(
    "#{event.user.mention} broke your stuff: #{exception}"
  )
end

bot.command(:test, rescue: rescue_proc) { 1 / 0 }
```

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -126,7 +126,10 @@ module Discordrb::Commands
         quote_start: attributes[:quote_start] || '"',
 
         # Quoted mode ending character
-        quote_end: attributes[:quote_end] || attributes[:quote_start] || '"'
+        quote_end: attributes[:quote_end] || attributes[:quote_start] || '"',
+
+        # Default block for handling internal exceptions, or a string to respond with
+        rescue: attributes[:rescue]
       }
 
       @permissions = {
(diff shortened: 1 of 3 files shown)


LikeLakers2:
I do wonder if this would be better as a `CommandBot` attribute rather than a `Command` attribute. We might also want to have it for both, with the `Command` attribute overriding the `CommandBot` attribute if there is one.

z64:
Yep, I was going to do this. Hadn't gotten around to it yet.

z64:
I think this is OK now, if the way I implemented the `CommandContainer#command` default attribute looks acceptable @meew0.

LikeLakers2:
It's been a while since I suggested adding it to CommandBot, and I'm wondering if the rescue attribute would fit better in `Bot` instead, given that you can give it a simple string. Just a suggestion, don't need to move it.

z64:
Yeah I remember. I don't think it's fitting right now to do so. Maybe later.

PR #362, opened 2017-06-14
Adding a simple send `hello world` message to a channel

awls99:
There's no simple hello world example for simple sending of messages, thought it might help other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,16 @@ bot.run
 
 This bot responds to every "Ping!" with a "Pong!".
 
+Or you can simply send a message to your chanel like this:
+
+```
+require 'discordrb'
+
+bot = Discordrb::Bot.new token: '<token>', client_id: 168123456789123456
+bot.run true
+bot.send_message  '123456789012345678', 'Hello world'
+```` 
+
 ## Support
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,6 +99,16 @@ bot.run
 
 This bot responds to every "Ping!" with a "Pong!".
 
+Or you can simply send a message to your chanel like this:
+
+```ruby
+require 'discordrb'
+
+bot = Discordrb::Bot.new token: '<token>', client_id: 168123456789123456
+bot.run true
```
z64: Running `async` here is pointless and doesn't really have much to do with this example. I would rather not encourage people to use this if they don't understand what they're doing. Please remove `true` and move it to the end of the file.

VxJasonxV:
Is this better suited as an [example file](https://github.com/meew0/discordrb/tree/master/examples)? The README is illustrating a series of common features (bot initialization, eventing, responding), but isn't a way to teach you how to use the library.

README.md
awls99: Fair enough, I was using it because I wanted to post a message and stop.

z64: Right - in which case, you can just completely omit `bot.run`. `send_message` just makes a REST call and does not need to initiate a gateway connection.

awls99:
>If you would like to add a send_message.rb there, and include other examples of sending messages (i.e. Channel#send_message, etc.) then this might be worth having.

Yes this, would be better in the examples, but  it would imply me knowing how, if I ever get around that I will add examples then

PR #368, opened 2017-06-25
Receive voice

Snazzah:
For now you can get opus data with `Listener#opus` and PCM data with `Listener#pcm`.
You can create a listener with `VoiceBot#listener`.
This PR is tested and does work but I think there should be something that can let the listener write to a file.

P.S. There is a random VWS error when the voice bot is destroyed with no error given.

- [ ] Write a users voice to a file
- [ ] Add meew's [dca1fixer](https://gist.github.com/meew0/cd2d2404c3e365ad1e71fc3520f582b0)

### DIFF
--- a/lib/discordrb/voice/encoder.rb
+++ b/lib/discordrb/voice/encoder.rb
@@ -103,4 +103,24 @@ module Discordrb::Voice
       @use_avconv ? "-vol #{(@filter_volume * 256).ceil}" : "-af volume=#{@filter_volume}"
     end
   end
+
+  # This  class decodes audio sent by Discord via UDP.
+  class Decoder
+    # Create a new decoder
+    def initialize
+      @sample_rate = 48_000
+      @frame_size = 960
+      @channels = 2
+
+      raise LoadError, 'Opus unavailable - voice not supported! Please install opus for voice support to work.' unless OPUS_AVAILABLE
+      @opus = Opus::Decoder.new(@sample_rate, @frame_size, @channels)
+    end
+
+    # Decodes the given buffer using opus.
+    # @param buffer [String] An encoded opus buffer.
+    # @return [String] A decoded opus buffer.
+    def decode(buffer)
+      @opus.decode(buffer)
+    end
+  end
 end
--- a/lib/discordrb/voice/encoder.rb
+++ b/lib/discordrb/voice/encoder.rb
@@ -104,7 +104,7 @@ module Discordrb::Voice
     end
   end
 
-  # This  class decodes audio sent by Discord via UDP.
+  # This class decodes audio sent by Discord via UDP.
   class Decoder
     # Create a new decoder
     def initialize
--- a/lib/discordrb/voice/listener.rb
+++ b/lib/discordrb/voice/listener.rb
@@ -64,7 +64,7 @@ module Discordrb::Voice
 
     def decode_packet(message, user)
       nonce = ([0] * 24)
-      for i in 0..11
+      (0..11).each do |i|
         nonce[i] = message[i]
       end
       message.slice!(0..11)
--- a/lib/discordrb/voice/listener.rb
+++ b/lib/discordrb/voice/listener.rb
@@ -41,20 +41,18 @@ module Discordrb::Voice
     # @note For internal use only
     # @visibility private
     def handle_packet(message)
-      Discordrb::LOGGER.debug "Received UDP packet"
+      Discordrb::LOGGER.debug 'Received UDP packet'
       ssrc = (message[8] * 0x1000000) + ((message[8 + 1] << 16) | (message[8 + 2] << 8) | message[8 + 3]).to_s.to_i.abs
       Discordrb::LOGGER.debug "Found packet by SSRC #{ssrc}"
       user = @ws.users[ssrc]
       if user
-      	Discordrb::LOGGER.debug "Identified packet by #{user.username}##{user.discriminator} (#{user.id})"
-      	if @queue[ssrc].nil?
-      	  decode_packet(message, user)
-      	else
-      	  @queue[ssrc].push message
-      	  @queue[ssrc].each do |m|
-      	    decode_packet(m, user)
-      	  end
-      	  @queue.delete ssrc
+        Discordrb::LOGGER.debug "Identified packet by #{user.username}##{user.discriminator} (#{user.id})"
+        if @queue[ssrc].nil?
+          decode_packet(message, user)
+        else
+          @queue[ssrc].push message
+          @queue[ssrc].each { |m| decode_packet(m, user) }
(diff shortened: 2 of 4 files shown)


Snazzah:
Hello i'm not dead but i need some way to put in the dca1 fixer so people can write working dca files

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+    def pcm(&block)
+      @pcm_blocks.push(block)
+    end
+
+    # Handles packets received by the UDP connection.
+    # @note For internal use only
+    # @visibility private
+    def handle_packet(message)
+      Discordrb::LOGGER.debug 'Received UDP packet'
+      ssrc = (message[8] * 0x1000000) + ((message[8 + 1] << 16) | (message[8 + 2] << 8) | message[8 + 3]).to_s.to_i.abs
```
z64: `.to_s.to_i.abs` ???

It looks like you're dangerously casting a signed binary value or something.. :cold_sweat: 
Can we refactor this and decode properly please?

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+      @opus_blocks.each do |block|
+        block.call(data, user)
+      end
+      pcm = @decoder.decode(data)
+      @pcm_blocks.each do |block|
+        block.call(pcm, user)
+      end
+    rescue RbNaCl::CryptoError => e
+      Discordrb::LOGGER.warn 'Failed to decrypt voice packet'
+      Discordrb::LOGGER.warn "Reason: #{e}"
```
z64: Please use `log_exception(e)`. Both of these should be `error`, not `warn`.

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def decode_packet(message, user)
+      nonce = ([0] * 24)
+      (0..11).each do |i|
+        nonce[i] = message[i]
+      end
+      message.slice!(0..11)
+      Discordrb::LOGGER.debug "Decoding packet by #{user.username}##{user.discriminator} (#{user.id})"
+      raise 'No secret key found!' unless @udp.secret_key
+      box = RbNaCl::SecretBox.new(@udp.secret_key)
```
z64: Does `@udp.secret_key` ever change in the lifetime of *this* object? At a glance I don't see why we have to do this instantiation for every packet. If you don't know when it's available, you can create a lazy getter method over an ivar.

lib/discordrb/voice/encoder.rb
```diff
@@ -103,4 +103,24 @@ def filter_volume_argument
       @use_avconv ? "-vol #{(@filter_volume * 256).ceil}" : "-af volume=#{@filter_volume}"
     end
   end
+
+  # This  class decodes audio sent by Discord via UDP.
+  class Decoder
```
z64: Logistically, I don't know why this is in `encoder.rb`

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      @udp.start_thread self
+    end
+
+    # Adds a handler for received opus data.
+    # @yield The block is executed when a packet has been received.
+    # @yieldparam data [String] The opus buffer.
+    # @yieldparam user [User] The user that is sending the buffer (is speaking).
+    def opus(&block)
+      @opus_blocks.push(block)
```
z64: `@opus_handlers` would be a better name for this. It's confusing to think that this might be storing "blocks" of opus data.

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    # Adds a handler for received PCM data.
+    # @yield The block is executed when a packet has been received.
+    # @yieldparam data [String] The decoded buffer.
+    # @yieldparam user [User] The user that is sending the buffer (is speaking).
+    def pcm(&block)
+      @pcm_blocks.push(block)
```
z64: ditto

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+      @queue = {}
+      @opus_blocks = []
+      @pcm_blocks = []
+
+      @udp.start_thread self
+    end
+
+    # Adds a handler for received opus data.
+    # @yield The block is executed when a packet has been received.
+    # @yieldparam data [String] The opus buffer.
```
z64: A string isn't a buffer :confused:

lib/discordrb/voice/listener.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    end
+
+    private
+
+    def decode_packet(message, user)
+      nonce = ([0] * 24)
+      (0..11).each do |i|
+        nonce[i] = message[i]
+      end
```
z64: This nonce algorithm seems unsafe / overkill. Can't we just unpack this?

lib/discordrb/voice/encoder.rb
Snazzah: Should I store it in a different file?

lib/discordrb/voice/listener.rb
Snazzah: I'm not actually sure how I did this so I probably can't refactor it

z64: Okay. I'll study the protocol over the next day or so to better help with this PR.

lib/discordrb/voice/listener.rb
z64: I think that `s/buffer/frame` is more accurate? When I read buffer I'm expecting something IO-like that can be `#read` from.

Snazzah: So should I convert it to an IO?

z64: I don't think so, no.

z64:
Discord's voice receive API is a bit of an undocumented mess, with staff saying its not guaranteed to be stable for bots until it is documented, and other libraries dealing with silent breaking changes. Because of that, and since no user has requested need for this, I'm going to close this, but I'm happy to revisit this implementation once there is reliable documentation.

Thank you for your efforts!

VladPutinEu:
> cumented mess, with staff saying its not guaranteed to be stable for bots until it is documented, and other libraries dealing with silent breaking changes. Because of that, and since no user has requested need for this, I'm going to close this, but I'm happy to revisit this implementation once there is reliable documentation.
> 
> Thank you for your efforts!

It's been a long time since you closed this, but I definitely see the need to add this. You are only one feature away from supporting all the features in the [unofficial Discord library comparison.](https://discordapi.com/unofficial/comparison.html) This would mean that this library would classify as top-tier.

PR #372, opened 2017-07-03
Fix Server#delete_role, Channel#delete_message

z64:
- `Server#delete role` was broken because the API method always expected a `reason` argument.
- Made all API methods that take a reason consistently nilable / optional so that adding reasons in the next release isn't a breaking change
- ~~Because `X-Audit-Log-Reason` is always handed off to `API.request` even if its value is `nil` (since it's an optional argument, this *actually* creates an unsightly entry in the audit log (see screencap). I've added a piece of logic to reject this header if it has a `nil` value so that this doesn't happen.~~
- `Channel#delete_message` was broken because its API method does not accept a `reason` argument, as it isn't auditable with a reason.
image

### DIFF
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -16,7 +16,7 @@ module Discordrb::API::Channel
 
   # Update a channel's data
   # https://discordapp.com/developers/docs/resources/channel#modify-channel
-  def update(token, channel_id, name, topic, position, bitrate, user_limit, reason)
+  def update(token, channel_id, name, topic, position, bitrate, user_limit, reason = nil)
     Discordrb::API.request(
       :channels_cid,
       channel_id,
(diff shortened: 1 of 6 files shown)


lib/discordrb/api.rb
```diff
@@ -91,6 +91,9 @@ def request(key, major_parameter, type, *attributes)
     # Add a custom user agent
     attributes.last[:user_agent] = user_agent if attributes.last.is_a? Hash
 
+    # Don't send a `nil` valued audit log reason header
+    attributes.last.delete('X-Audit-Log-Reason') if attributes.last['X-Audit-Log-Reason'].nil?
```
meew0: The unsightly entry is because the reason is always given as a keyword argument even if that would be wrong, so it's a hash at times. (This is kind of my fault for not noticing that before, sorry.) The fix is to change all the reason parameters in API method calls to only be keyword arguments if necessary.

Regardless of that, however, this statement does nothing – the `X-Audit-Log-Reason` key is given as a symbol in method calls and treated as a string here. RestClient appears to drop nil header values, or at least there was no reason in the audit logs during my testing with the fix I mentioned above applied, unless I specified one.

PR #380, opened 2017-07-28
Implement Channel#nsfw

z64:


z64:
@Daniel-Worrall is there anything reasonable I can spec out of this? I was thinking the setter, `#nsfw=` but it seems kinda notworth atm as it would requiring stubbing `#update_data`, which i believe I'd want to set up a fixture and stuff up for it for other methods.. if I'm understanding this right.

z64:
Perhaps this is sufficient for now.

PR #382, opened 2017-07-31
Update concept of "default channel" per API change

z64:
Discord has (pending deployment at time of writing this PR) updated their API to remove the previous concept of the "default channel" being the channel with the same snowflake ID as the guild. It is now described as:

> ".. the channel with the highest position that their (the client) allows them to see (read)".

This API change also allows "default channels" to be deleted and hidden just like any other channel.

https://github.com/hammerandchisel/discord-api-docs/pull/329

### DIFF
--- a/lib/discordrb/data.rb
+++ b/lib/discordrb/data.rb
@@ -2667,9 +2667,18 @@ module Discordrb
       @owner = member(@owner_id) if exists
     end
 
-    # @return [Channel] The default channel on this server (usually called #general)
+    # The default channel is the text channel on this server with the highest position
+    # that the client has Read Messages permission on.
+    # @return [Channel] The default channel on this server
     def default_channel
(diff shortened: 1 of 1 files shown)


z64:
Woops this isn't quite right. Have to rework to scan channel's permission overwrites for @-everyone.

z64:
This looks like it works properly.

z64:
Squashed my fixup - this change is now fully live.

LikeLakers2:
I know that this is a closed PR now, but because of the API change, it's now possible to have *no* channels whatsoever in a server. While there's little point to having a completely blank server, and even less point to having a bot join one, I'd like to ask that it be made sure that Discordrb can cope with a lack of channels in a server.

meew0:
I tested it a bit and noticed no problems, so it should be fine.

PR #387, opened 2017-08-05
Make sure to always ratelimit preemptively, fixes #378

### LINKED ISSUE #378: Some requests fail to be preemptively rate limited
Running this produces a 429:
```rb
> 2.times { BOT.user(1) }
2017-07-19 16:50:00.529 main R Locking RL mutex (key: [:users_uid, nil]) for 0.928 seconds due to Discord rate limiting
```
Headers are present on the route:
```
x-ratelimit-limit →1
x-ratelimit-remaining →0
x-ratelimit-reset →1500483223
```
Body:
```json
{
    "global": false,
    "message": "You are being rate limited.",
    "retry_after": 278
}
```
I'm unsure how this gets by [this line](https://github.com/meew0/discordrb/blob/master/lib/discordrb/api.rb#L109).

I think this is specifically to do with 1/1 routes, but reactions are preemptively locked "correctly" so.. I'm not sure.

unleashy:
So basically, issue #378 happens because the `#user` method has this block of code in it:
```ruby
begin
  response = API::User.resolve(token, id)
rescue RestClient::ResourceNotFound
  return nil
end
```
As you can see, the `rescue` block makes sure to return `nil` in case a user is not found. However, this is thrown by RestClient. In `#request`, we have this:
```ruby
response = raw_request(type, attributes)

if response.headers[:x_ratelimit_remaining] == '0' && !mutex.locked?
  # handles preemptive ratelimiting
end
```
All `raw_request` does is delegate to RestClient for the request, and when RestClient gets a 404 it raises a `ResourceNotFound` exception which is then caught by the rescue block in #user, thus skipping the preemptive ratelimiting completely, and since #user calls a 1/1 route, you get b1nzy'd.

All I did was wrap the `#raw_request` call into an `begin-ensure` block that calls `#handle_preemptive_rl` that does the preemptive ratelimiting, so this makes sure that the preemptive ratelimiting handling is _always_ ran no matter what. This seems to fix the issue.

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -114,7 +114,6 @@ module Discordrb::API
         if response
           handle_preemptive_rl(response, mutex, key) if response.headers[:x_ratelimit_remaining] == '0' && !mutex.locked?
         else
-          # technically this should never happen... but you know.
           Discordrb::LOGGER.ratelimit('Response was nil when trying to preemptively rate limit!!')
         end
       end
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -140,7 +140,7 @@ module Discordrb::API
   end
 
   def handle_preemptive_rl(response, mutex, key)
-    Discordrb::LOGGER.debug "RL bucket depletion detected! Date: #{response.headers[:date]} Reset: #{response.headers[:x_ratelimit_reset]}"
+    Discordrb::LOGGER.ratelimit "RL bucket depletion detected! Date: #{response.headers[:date]} Reset: #{response.headers[:x_ratelimit_reset]}"
 
     now = Time.rfc2822(response.headers[:date])
     reset = Time.at(response.headers[:x_ratelimit_reset].to_i)
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -104,20 +104,19 @@ module Discordrb::API
       # If the global mutex happens to be locked right now, wait for that as well.
       mutex_wait(@global_mutex) if @global_mutex.locked?
 
-      response = raw_request(type, attributes)
-
-      if response.headers[:x_ratelimit_remaining] == '0' && !mutex.locked?
-        Discordrb::LOGGER.ratelimit "RL bucket depletion detected! Date: #{response.headers[:date]} Reset: #{response.headers[:x_ratelimit_reset]}"
-
-        now = Time.rfc2822(response.headers[:date])
-        reset = Time.at(response.headers[:x_ratelimit_reset].to_i)
-
-        delta = reset - now
-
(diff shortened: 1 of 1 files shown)


lib/discordrb/api.rb
```diff
@@ -104,20 +104,19 @@ def request(key, major_parameter, type, *attributes)
...
+      begin
+        response = raw_request(type, attributes)
+      rescue RestClient::Exception => e
+        response = e.response
+        raise e
+      ensure
+        if response
+          handle_preemptive_rl(response, mutex, key) if response.headers[:x_ratelimit_remaining] == '0' && !mutex.locked?
+        else
+          # technically this should never happen... but you know.
```
z64: I don't think this comment needs to be here.

lib/discordrb/api.rb
```diff
@@ -140,6 +139,18 @@ def request(key, major_parameter, type, *attributes)
     response
   end
 
+  def handle_preemptive_rl(response, mutex, key)
```
z64: Could you add a docstring for this? At least a basic one like `#request` has, but feel free to go into more detail.

lib/discordrb/api.rb
```diff
@@ -140,6 +139,18 @@ def request(key, major_parameter, type, *attributes)
...
+    Discordrb::LOGGER.debug "RL bucket depletion detected! Date: #{response.headers[:date]} Reset: #{response.headers[:x_ratelimit_reset]}"
```
z64: Why did you move this to `LOGGER.debug` instead of `LOGGER.ratelimit`?

lib/discordrb/api.rb
```diff
@@ -104,20 +104,19 @@ def request(key, major_parameter, type, *attributes)
...
+        response = raw_request(type, attributes)
+      rescue RestClient::Exception => e
+        response = e.response
+        raise e
+      ensure
+        if response
+          handle_preemptive_rl(response, mutex, key) if response.headers[:x_ratelimit_remaining] == '0' && !mutex.locked?
+        else
+          # technically this should never happen... but you know.
+          Discordrb::LOGGER.ratelimit('Response was nil when trying to preemptively rate limit!!')
```
z64: I'd prefer: "Response was nil before trying to preemptively rate limit!"

lib/discordrb/api.rb
unleashy: Wtf. I don't remember changing this...

lib/discordrb/api.rb
unleashy: Tbh this should be a private function, but I couldn't get it to be private for some reason. But I'll add a doc

z64: Since these methods succeed a `module_function` declaration, they are class methods.

https://stackoverflow.com/questions/4952980/how-to-create-a-private-class-method

It's best left public though, I think, at least consistent with the other API wrapper internals like `#raw_request`.

PR #391, opened 2017-08-08
Cache channels before updating server data, fixes #374

### LINKED ISSUE #374: Avoid GET-ing hidden channels, don't raise exceptions
The bot sometimes attempts to `GET` API information about a hidden channel, and it gets a 403 from Discord with "No Access". Then it raises a `Discordrb::Errors::NoPermission` and continues. This is harmless behavior, but it leads to a lot of annoyances developing with the bot:

- If you use an exception coalition service, it fills up with noise about these exceptions
- It's not obvious why you got this exception at first glance (see #373)
- If you hook into this error workflow, you will have to handle a lot of raised exceptions that are harmless

Raising and catching exceptions for semi-expected behavior like this is not a good workflow. Ruby exceptions are very expensive to toss.

I'm not sure about the difficulty of this but one of these would be more ideal:

- Avoid trying to `GET` a hidden channel beforehand somehow (not sure how you'd know the channel was hidden)
- RestClient supports a [block-style syntax](https://github.com/rest-client/rest-client#response-callbacks-error-handling) for methods which lets you inspect all parts of the process and doesn't raise exceptions on error codes. When you encounter a hidden channel, log a message but don't raise an exception.

z64:
This fixes a bug where channels we don't have access to (because we can't Read Messages or Connect) are attempted to be resolved by `GET /channels/{cid}`, when in fact, we already have the data for these channels in `GUILD_CREATE` and can properly cache them from there.

For example, this would cause exceptions to be raised every time a user joins or leaves (voice state update) a voice channel that denies Connect to our client.

For more information on this behavior, see https://github.com/meew0/discordrb/issues/374#issuecomment-313839853

The logic in `update_data` that refers to a channel being "unavailable" and setting AFK channels to nil can probably be removed, and maybe the embed_channel as well, but I need to look into this more closely to tidy it up. Opening this PR as a placeholder in the meantime.

---

Todo:
- [x] Move `update_data` back up, but place `process_channels` before it to isolate any future edge cases
- [x] Remove "unavailable channel" exception check for `@afk_channel`
- [x] Remove "unavailable channel" exception check for `@embed_channel`
- [x] Fix `||` logic in some `update_data` statements

### DIFF
--- a/lib/discordrb/data.rb
+++ b/lib/discordrb/data.rb
@@ -2683,17 +2683,17 @@ module Discordrb
     end
 
     # @return [true, false] whether or not the server has widget enabled
-    def embed?
-      update_data if @embed.nil?
-      @embed
+    def embed_enabled?
+      update_data if @embed_enabled.nil?
+      @embed_enabled
     end
-    alias_method :widget_enabled, :embed?
-    alias_method :widget?, :embed?
-    alias_method :embed_enabled, :embed?
+    alias_method :widget_enabled, :embed_enabled?
+    alias_method :widget?, :embed_enabled?
+    alias_method :embed?, :embed_enabled?
 
     # @return [Channel, nil] the channel the server embed will make a invite for.
     def embed_channel
-      update_data if @embed.nil?
+      update_data if @embed_enabled.nil?
       @bot.channel(@embed_channel_id) if @embed_channel_id
     end
     alias_method :widget_channel, :embed_channel
@@ -2747,8 +2747,8 @@ module Discordrb
     # @return [String, nil] the widget URL to the server that displays the amount of online members in a
     #   stylish way. `nil` if the widget is not enabled.
     def widget_url
-      update_data if @embed.nil?
-      return nil unless @embed
+      update_data if @embed_enabled.nil?
+      return unless @embed_enabled
       API.widget_url(@id)
     end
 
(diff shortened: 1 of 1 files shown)


meew0:
Are you sure this doesn't cause any problems with some of the other `process_` methods requiring server data to be present? From a quick look I haven't seen any problems but there's a lot of code in e.g. initializers I might have overlooked.

z64:
As we discussed in Discord I moved the location of `process_channels` / `update_data`, and removed the "unavailability" handling that should never occur.

Also, see my comments in the full commit message of 2baf63a. I updated behavior of `Server#update_data` to remove the need for `cache_embed`. Just to recap, it was necessary to resolve the server as `embed_channel_id`, as `GUILD_CREATE` doesn't provide `embed_channel_id`, and we may have not otherwise received an update yet to the guild that sets that attribute. We were just picking out that attribute, but we may as well not "waste" the rest of the "fresh" data.

I tested all of the immediately affected methods and everything works fine. I may do some more testing later (leaving my bot on this branch for now too).

z64:
Before I forget too, I'll also fix the third clause in the `update_data` statements from not being used due to the second clause not being nilable in this PR, as we discussed briefly

z64:
My todo's are done here.

z64:
I'll note from 9e92f76, in a later update, we should instead use the REST calls specifically for guild embeds ( #435 ) instead of the whole server object.

PR #395, opened 2017-08-12
Much embed cleanup. So accessor. Wow.

LikeLakers2:
Just some cleanup for the embed classes.

- All attribute used in `EmbedAuthor`, `EmbedField`, `EmbedFooter`, `EmbedImage` and `EmbedThumbnail` are now exposed as attributes. No more `#instance_variable_get` and `#instance_variable_set` for the devs who wanted (or needed) to get/set them afterwards.
- Documentation, documentation, documentation. More specifically, I documented that most attributes in the above classes (as well as `Embed`) will be `nil` by default if not set when first instancing the class.
- `EmbedField`'s `inline` attribute now defaults to false, as Discord defaults it to false anyways. (Tested by forcing a hash without `inline` defined to be sent) This change should not affect existing bots except where they were checking if `inline` is explicitly `nil`.
- The `fields` attribute within `Embed` is now an accessor rather than a reader. While we have `Embed#<<` and `Embed#add_field`, there is no reason we should be required to use them to add a field, especially when we allow setting `fields` to anything we want when instancing `Embed` anyways.
- Also apparently I fixed two Code Climate issues????

### DIFF
--- a/lib/discordrb/webhooks/embeds.rb
+++ b/lib/discordrb/webhooks/embeds.rb
@@ -19,22 +19,22 @@ module Discordrb::Webhooks
     end
 
     # The title of this embed that will be displayed above everything else.
-    # @return [String]
+    # @return [String, nil] title of the embed, or `nil` if none was set during initialization
     attr_accessor :title
 
     # The description for this embed.
-    # @return [String]
+    # @return [String, nil] description for this embed, or `nil` if none was set during initialization
     attr_accessor :description
 
     # The URL the title should point to.
-    # @return [String]
+    # @return [String, nil] URL the title should point to, or `nil` if none was set during initialization
     attr_accessor :url
 
     # The timestamp for this embed. Will be displayed just below the title.
-    # @return [Time]
+    # @return [Time, nil] timestamp for this embed, or `nil` if none was set during initialization
     attr_accessor :timestamp
 
-    # @return [Integer] the colour of the bar to the side, in decimal form.
+    # @return [Integer, nil] the colour of the bar to the side, in decimal form, or `nil` if none was set during initialization
     attr_reader :colour
     alias_method :color, :colour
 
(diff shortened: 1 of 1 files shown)


lib/discordrb/webhooks/embeds.rb
```diff
@@ -198,11 +226,23 @@ def to_hash
...
+    attr_accessor :inline
+
     # Creates a new field object.
-    # @param name [String, nil] The name of the field, displayed in bold at the top.
+    # @param name [String, nil] The name of the field, displayed in bold at the top of the field.
     # @param value [String, nil] The value of the field, displayed in normal text below the name.
-    # @param inline [true, false] Whether the field should be displayed in-line with other fields.
-    def initialize(name: nil, value: nil, inline: nil)
+    # @param inline [true, false] Whether the field should be displayed inline with other fields. Defaults to false.
+    def initialize(name: nil, value: nil, inline: false)
```
z64: I'm pretty conflicted about providing a default just for this field and nothing else. Passing `nil` will have the same effect, as you said, but sending this object as it is initialized will create a Bad Request anyways because of the other two arguments not being set. So ultimately this affects nothing. I don't see any reason to change this, but I don't have a reason to say no either. Just not sure why you chose to do this for this specific attribute, I guess.

lib/discordrb/webhooks/embeds.rb
```diff
@@ -198,11 +226,23 @@ def to_hash
...
+    # The name of the field, displayed in bold at the top of the field.
+    # @return [String, nil] name of the field, or `nil` if none was set during initialization
+    attr_accessor :name
+
+    # The value of the field, displayed in normal text below the name.
+    # @return [String, nil] value of the field, or `nil` if none was set during initialization
+    attr_accessor :value
+
+    # Whether the field should be displayed inline with other fields.
+    # @return [true, false] whether the field should be displayed inline with other fields. Defaults to false.
```
z64: You don't have to document that it defaults to `false`. This will be shown in the rendered documentation in the initializer decorator-thing. The same goes for `inline` below.

lib/discordrb/webhooks/embeds.rb
LikeLakers2: Because it felt like the better change to make compared to documenting that `@inline` might potentially be nil.

lib/discordrb/webhooks/embeds.rb
```diff
@@ -124,6 +124,14 @@ def to_hash
   # An embed's footer will be displayed at the very bottom of an embed, together with the timestamp. An icon URL can be
   # set together with some text to be displayed.
   class EmbedFooter
+    # The text to be displayed in the footer.
```
LikeLakers2: Will change. I wasn't sure whether to do this or not as this exact thing also happens in several other places within Discordrb as a whole. Still, will change.

lib/discordrb/webhooks/embeds.rb
```diff
@@ -19,22 +19,22 @@ def initialize(title: nil, description: nil, url: nil, timestamp: nil, colour: n
     end
 
     # The title of this embed that will be displayed above everything else.
-    # @return [String]
+    # @return [String, nil] title of the embed, or `nil` if none was set during initialization
```
LikeLakers2: I'd prefer to either be explicit, or set a default for these. Setting a default doesn't feel like a good solution. Do you really still think I should remove all those?

z64: I think it's unnecessary because users control the *entire* state of these objects. The library will never create them on their own, and user takes full responsibility for their state. If it is nil, the user knows this.

LikeLakers2: So just to be clear, do you want me to keep the `[T, nil]` and remove the `, or nil if none was set during initialization`? Or remove both?

z64: Yeah, keep the nilable type.

lib/discordrb/webhooks/embeds.rb
LikeLakers2: Also, I chose to avoid making defaults for anything else for fear that it might potentially break something. If you want me to set obvious defaults for everything else, however, I will gladly do so.

PR #400, opened 2017-08-19
Fix arg typo for Argument Error in CommandBot

Daniel-Worrall:
As noted by Ghosty in Discord, `type` is never defined.


TrenchFroast:
idea: define type as types[i], and then replace all occurrences of types[i] with type; this should in theory actually be faster

Daniel-Worrall:
This is fine as is.

PR #405, opened 2017-09-07
Update README.md

PixeLInc:
Update to add 'bundle exec' in front of running the ruby file so it will find discordrb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,13 @@ And then install via `bundle install`.
 
 Run the [ping example](https://github.com/meew0/discordrb/blob/master/examples/ping.rb) to verify that the installation works (make sure to replace the token and client ID in there with your bots'!):
 
+To run the bot while using bundler: 
+
     bundle exec ruby ping.rb
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,6 +65,10 @@ Alternatively, while Bundler is the recommended option, you can also install dis
 > **Make sure you have the DevKit installed! See the [Dependencies](https://github.com/meew0/discordrb#dependencies) section)**
 
     gem install discordrb --platform=ruby
+    
+To run the bot with Gem, unlike bundler you can just do:
```
z64: > unlike bundler

There's no need to draw comparisons between the two here. just "Run the bot by running:" or "Run the bot with:" is fine.

z64:
I'll also just note for fun; there is also the alternative of `require 'bundler/setup'` in your code, and then just `ruby bot.rb` will work. Technicalities like that is sorta why I was originally inclined to leave it out ; users should read bundler's docs / most are already probably familiar with bundler.

This is good n' fine though.

PR #412, opened 2017-09-21
ensure_user when initializing a message with an uncached author

z64:
In certain use cases like scraping long channel history, we may come across uncached users who have left the guild. We would make an API request here to resolve them, but a message author is always a full User object which we can cache with ensure_user instead of making a request to the API.

Also notably, this would cause exceptions on selfbots who cannot access `GET /users/{user.id}`

---
Todo:
- [x] Fix specs per Nekka's review

### DIFF
--- a/spec/data_spec.rb
+++ b/spec/data_spec.rb
@@ -67,6 +67,30 @@ module Discordrb
     end
   end
 
+  describe Message do
+    let(:server) { double('server') }
+    let(:channel) { double('channel', server: server) }
+    let(:token) { double('token') }
+    let(:bot) { double('bot', channel: channel, token: token) }
+
+    fixture :message_data, %i[message]
+    fixture_property :message_author, :message_data, ['author']
+
+    describe '#initialize' do
+      it 'caches an unavailable author' do
+        allow(server).to receive(:member)
+        allow(channel).to receive(:private?)
+        allow(channel).to receive(:text?)
+
+        # Bot will receive #ensure_user because the observed message author
+        # is not present in the server cache, which is possible
+        # (for example) if the author had left the server.
+        expect(bot).to receive(:ensure_user).with message_author
+        described_class.new(message_data, bot)
+      end
+    end
+  end
+
   describe Webhook do
     let(:token) { double('token') }
     let(:reason) { double('reason') }
--- a/lib/discordrb/data.rb
+++ b/lib/discordrb/data.rb
@@ -2291,7 +2291,7 @@ module Discordrb
 
                     unless member
                       Discordrb::LOGGER.debug("Member with ID #{data['author']['id']} not cached (possibly left the server).")
-                      member = @bot.user(data['author']['id'].to_i)
+                      member = @bot.ensure_user(data['author'])
                     end
 
                     member
--- a/spec/data/message.json
+++ b/spec/data/message.json
@@ -1,0 +1,21 @@
+{
+    "attachments": [],
+    "tts": false,
+    "embeds": [],
+    "timestamp": "2017-09-21T19:53:15.684000+00:00",
+    "mention_everyone": false,
+    "id": "360513832470315009",
+    "pinned": false,
+    "edited_timestamp": null,
+    "author": {
+        "username": "Ghosty",
+        "discriminator": "8204",
+        "id": "97437477723193344",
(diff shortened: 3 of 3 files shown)


spec/data_spec.rb
```diff
@@ -30,6 +32,19 @@ module Discordrb
...
     end
 
+    describe '#history' do
+      context 'when reading history with uncached authors not in the server' do
+        it 'caches the full user object from the full author' do
+          allow(API::Channel).to receive(:messages).and_return(history_json)
+          allow(bot).to receive(:channel).and_return(channel)
+          allow(server).to receive(:member).and_return(nil)
+
+          expect(bot).to receive(:ensure_user).at_least(1).times
```
z64: I should note, it was getting called two times and apparently `expect` defaults to only one time.
However, without looking very hard, I can't tell where ensure_user was getting called twice, even if I changed the below call to `#history(1)`. It passes anyway with `at_least`, I'm still curious though.

Daniel-Worrall: If this is the case, shouldn't this be `#exactly(1).times`?

z64: Technically it should be two times with the below `history(2)`, since I'm stubbing `ensure_user`'s behavior, both messages in the history JSON should never actually be cached. 

But, even if I only read *one* message, it would apparently call it twice..

z64: Sure. I still would like to understand why though, as it doesn't seem to be because there's two messages..

z64: just to close this sad train of comments: TL;DR - zac's dumb. It is because the stub always returns two messages. :smile:

spec/data_spec.rb
```diff
@@ -30,6 +32,19 @@ module Discordrb
       end
     end
 
+    describe '#history' do
+      context 'when reading history with uncached authors not in the server' do
+        it 'caches the full user object from the full author' do
+          allow(API::Channel).to receive(:messages).and_return(history_json)
+          allow(bot).to receive(:channel).and_return(channel)
+          allow(server).to receive(:member).and_return(nil)
```
Daniel-Worrall: Since you're expecting the external method of `ensure_user` to do the job of caching, we can put here `and_return(nil, double)` so that it'll always return a "cached user" after the first calls and we know it only calls this once per method. This also solves the below issue of `at_least`

spec/data_spec.rb
Daniel-Worrall: Okay, yeah, the stub always returns 2 messages so this should actually be written as `receive(:ensure_user).twice` (not `exactly(2).times` because English). In actuality, we shouldn't expect this method to receive it twice since caching happens, so we should rewrite the stubs to allow for that.

In ACTUAL actuality, this shouldn't even be under `#history`, and should actually be under `Message` as that's where the bug lies and that's what we're actually testing. More on that on review message.

z64:
@Daniel-Worrall I think this is G2G.

spec/data_spec.rb
```diff
@@ -55,6 +55,27 @@ module Discordrb
...
+    fixture_property :message_author, :message_data, ['author']
+
+    describe '#initialize' do
+      it 'caches an unavailable author' do
+        allow(server).to receive(:member)
+        allow(channel).to receive(:private?)
+        allow(channel).to receive(:text?)
+
+        expect(bot).to receive(:ensure_user).with message_author
+        described_class.new(message_data, bot)
```
Daniel-Worrall: Comment here for clarity

z64:
It is essentially speccing this warning in the immediate line above:
```rb
Discordrb::LOGGER.debug("Member with ID #{data['author']['id']} not cached (possibly left the server).")
member = @bot.ensure_user(data['author'])
```
https://github.com/z64/discordrb/blob/188fa9fd61f759ba01c68504ba66de2a01f8c4ae/lib/discordrb/data.rb#L2176

It's because members that have left all mutual guilds (not READY..?) will not be cached, and calls like `Channel#history` will yield a case where `member` in that `unless` branch is `nil`, but we can still produce a `User` since `author` is always a full user object.

I can make a comment to the above ^ effect then?

Daniel-Worrall:
I guess, but I just meant the reason that it triggers the call because it has no cached users so it'll happen no matter what data you receive

PR #416, opened 2017-09-26
Implement a more useful way of sorting channels

meew0:
Depends on #415.

Adds a method `Channel#sort_after` that can be used to sort a channel after another in the list. Uses the same algorithm as Discord so it should be fairly reliable.

### DIFF
--- a/lib/discordrb/api/channel.rb
+++ b/lib/discordrb/api/channel.rb
@@ -16,13 +16,13 @@ module Discordrb::API::Channel
 
   # Update a channel's data
   # https://discordapp.com/developers/docs/resources/channel#modify-channel
-  def update(token, channel_id, name, topic, position, bitrate, user_limit, nsfw, reason = nil)
+  def update(token, channel_id, name, topic, position, bitrate, user_limit, nsfw, permission_overwrites, parent_id, reason = nil)
     Discordrb::API.request(
       :channels_cid,
       channel_id,
       :patch,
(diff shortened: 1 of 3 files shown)


z64:
Took some time to fix the history of this PR and review this implementation.

This works except for what seems to be an edge case where I tried to insert a channel into the last position of a category (after the last channel) and it actually wound up sorting above it, similar to what happens if you set two channels to the same `position` (this is "undefined behavior" by Discord, unlike roles). I can probably reproduce this again and look more closely into why this happens.

But moreover, this seemed a little over-complicated; I took an alternative approach of getting all the channels within the context of `other`, inserting `self`'s ID into the appropriate position in a prepared array, and then building `move_argument`'s `position` by simply doing `each_with_index` so that it re-indexes position from `0`. This should guarantee that every channel will be sorted correctly, without leaving any channel behind.

https://github.com/z64/discordrb/blob/8972d84b222bcdcd027e4f81e8dd9e71a0e57aaa/lib/discordrb/data.rb#L1374-L1434

Implementation details aside, I think that some people may try to sort a voice channel after the last text channel in a category, as you "visually" do in the desktop client. I guess what I mean is, we should more clearly document the channel type matching exception & discord's behavior. (I don't think it's worth the effort to detect if we're sorting after *"the last text channel"*)

z64:
Closed in favor of #497

PR #419, opened 2017-09-30
Add missing error codes to Discordrb::Errors

LikeLakers2:
Adds JSON response code errors from the Discord API docs that were missing.

Also renames `InvalidForDM` to `InvalidActionForDM` so we're being explicit.


z64:
Thanks!

PR #420, opened 2017-10-03
Remove `mentions` in API::Channel#send_message, Respondable tweaks

z64:
- Removes `mentions` from `API::Channel#send_message`. Unsure where this came from, but Discord doesn't support sending this; it is just ignored, and we always sent a hard-coded empty array anyways in our abstraction.
- Removes `MessageEvent#send_message`. This was already included from `Respondable`, and was being overwritten.
- Adds support for `embed`s arguments in `Respondable#send_message`, and adds `Respondable#send_embed` forwarding to `Channel#send_embed`


meew0:
The first thing is because Discord *used* to support this; you could e.g. send a message that contains a bunch of mentions, then set `mentions` to an empty array so the mentions would be resolved but nobody would actually be pinged.

PR #427, opened 2017-10-27
Add `increment` parameter to `Bucket#rate_limited?`.

Lavode:
This adds the ability to increment the internal counter by an arbitrary
amount. This allows to eg properly rate-limit commands which can have
varying impact on other systems.

Note that there is a YARD warning regarding the doc string on line 40:
```
[warn]: @param tag has unknown parameter name: increment: 
    in file `lib/discordrb/commands/rate_limiter.rb' near line 42
```

However, the documentation is generated correctly. Changing `increment:` to `increment` in the docs silences the warning, but leads to YARD not picking up the default value.

### DIFF
--- a/lib/discordrb/commands/rate_limiter.rb
+++ b/lib/discordrb/commands/rate_limiter.rb
@@ -37,8 +37,9 @@ module Discordrb::Commands
     # Performs a rate limiting request
     # @param thing [#resolve_id, Integer, Symbol] The particular thing that should be rate-limited (usually a user/channel, but you can also choose arbitrary integers or symbols)
     # @param rate_limit_time [Time] The time to base the rate limiting on, only useful for testing.
+    # @param increment [Integer] How much to increment the rate-limit counter. Default is 1.
     # @return [Integer, false] the waiting time until the next request, in seconds, or false if the request succeeded
-    def rate_limited?(thing, rate_limit_time = nil)
+    def rate_limited?(thing, rate_limit_time = nil, increment: 1)
       key = resolve_key thing
       limit_hash = @bucket[key]
 
@@ -47,7 +48,7 @@ module Discordrb::Commands
         @bucket[key] = {
           last_time: Time.now,
           set_time: Time.now,
-          count: 1
+          count: increment
         }
 
         return false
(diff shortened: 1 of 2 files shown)


z64:
This looks fine, however;

> This allows to eg properly rate-limit commands which can have varying impact on other systems.

Could you please provide a more detailed use case? I'm not entirely sure why one would use this, where it seems you could already accomplish something similar just by using multiple buckets, or just calling `#rate_limited?` multiple times programatically with an equivalent rate limit.

lib/discordrb/commands/rate_limiter.rb
```diff
@@ -37,8 +37,9 @@ def clean(rate_limit_time = nil)
     # Performs a rate limiting request
     # @param thing [#resolve_id, Integer, Symbol] The particular thing that should be rate-limited (usually a user/channel, but you can also choose arbitrary integers or symbols)
     # @param rate_limit_time [Time] The time to base the rate limiting on, only useful for testing.
+    # @param increment: [Integer] How much to increment the rate-limit counter.
```
z64: I would prefer not to generate YARD warnings. Consistent with the rest of the library, you can just write `"Default is 1"` at the end of the docstring.

Lavode:
Fair point, the commit message is not really a great description. ;)

I have a command `!excuse <count = 1>`, which makes the bot spit out `count` excuses, each on its own line. I want to rate limit the number of lines per time interval which a user can generate this way, in order not to generate too much noise.

Clearly, `!excuse 1`, which generates 1 line of output, and `!excuse 10`, which generates 10 lines, have to be treated differently rate-limiting wise.

As `!excuse 1` and `!excuse 5` should contribute towards the same (per-user) rate limit, using multiple buckets is not really an option. (Not to mention that it's a bit impractical to generate one bucket for each possible count.)

Calling `#rate_limited?` multiple times will not work as I want the command to either be processed completely, or not at all, rather than aborting midway through due to hitting the rate limit.

In a similar manner, other use cases might include commands which allow to call foreign APIs 1 to n times, or commands which allow to do computation with various complexities (eg approximation with varying iteration counts), and so on.

lib/discordrb/commands/rate_limiter.rb
Lavode: Sure, can do. Do you prefer a separate, or squashed, commit?

z64: You can squash it into your existing commit, thanks! :+1:

Issue #430, opened 2017-11-15
Awaits added to Users trigger twice.


W0lf619:
I was testing code the example for await to understand how it worked, specifically the number guessing game, and discovered that the await event would fire twice each time false was returned. So I would enter a number, and the bot would respond with it being either too high or too low twice, until I guessed correctly.

Here is a code snippet of what I was testing:
```ruby
# Start the game by typing "!game" in chat.
bot.message(start_with: '!game') do |event|
  # Pick a number between 1 and 10
  magic = rand(1..10)

  # Await a MessageEvent specifically from the invoking user.
  #
  # Note that since the identifier I'm using here is `:guess`,
  # only one person can be playing at one time. You can otherwise
  # interpolate something into a symbol to have multiple awaits
  # for this "command" available at the same time.
  event.user.await(:guess) do |guess_event|
    # Their message is a string - cast it to an integer
    guess = guess_event.message.content.to_i

    # If the block returns anything that *isn't* `false`, then the
    # event handler will persist and continue to handle messages.
    if guess == magic
      # This returns `nil`, which will destroy the await so we don't reply anymore
      guess_event.respond 'you win!'
    else
      # Let the user know if they guessed too high or low.
      guess_event.respond(guess > magic ? 'too high' : 'too low')

      # Return false so the await is not destroyed, and we continue to listen
      false
    end
  end

  # Let the user know we're  ready and listening..
  event.respond 'Guess a number between 1 and 10..'
end
```

z64:
Hi there @W0lf619 !

Sorry - I can't reproduce this; copy and pasted the code from your issue.

![image](https://user-images.githubusercontent.com/6119032/32815456-2d4b50c0-c982-11e7-8739-f250993a4908.png)

Are you sure you weren't running two instances of the bot by accident?

W0lf619:
I'm sure. Though I believe I've figured out the conditions that causes it to send twice. I test my commands by PMing my bot. Seems that if the bot is listening for an AwaitEvent, that event registers twice if it's in a PM.

![await bug](https://user-images.githubusercontent.com/32836691/32832605-f3a970c8-c9f3-11e7-9fcb-1bb79e0af274.png)

Outside of a PM, in a regular channel on a server, the event is only registered once, as your testing shows.

z64:
Thank you for reporting. This makes sense as to why it's happening, but I'm very unsure how one would fix this. We'll take a closer look into it. :+1: May be slightly related to #422

[...]

PR #431, opened 2017-11-15
Only check for matching events of the correct class in awaits

### LINKED ISSUE #422: Inheritance breaks certain event handler behavior
```rb
bot.add_await(:foo, Discordrb::Events::ReactionAddEvent) do
 # ...
end
```
The above code will be raised on *both* `ReactionAddEvent` *and* `ReactionRemoveEvent` because:

<https://github.com/meew0/discordrb/blob/master/lib/discordrb/bot.rb#L1238>
<https://github.com/meew0/discordrb/blob/master/lib/discordrb/events/generic.rb#L92>
<https://github.com/meew0/discordrb/blob/master/lib/discordrb/events/reactions.rb#L53>

```rb
class ReactionEvent
end

class ReactionAddEvent < ReactionEvent
end

class ReactionRemoveEvent < ReactionEvent
end

# When #matches? is called for checking the await handlers against the event..
ReactionAddEvent.new.is_a?(ReactionEvent) # true
ReactionRemoveEvent.new.is_a?(ReactionEvent) # true
```
As a workaround you can check for the specific event class yourself.

There may be other cases of this, but I haven't looked. This was asked about in the channel a while back ago and I tested it.

z64:
When processing awaits, `#matches?` would incorrectly return true for *different* events that share a superclass, meaning some events would trigger awaits twice, or be triggered on different sibling events.

This checks up front, before processing the awaited event, that the event is an instance of the specific event class being awaited, ignoring inheritance.

Closes #422, #430

### DIFF
--- a/lib/discordrb/await.rb
+++ b/lib/discordrb/await.rb
@@ -41,7 +41,7 @@ module Discordrb
     # @return [Array] This await's key and whether or not it should be deleted. If there was no match, both are nil.
     def match(event)
       dummy_handler = EventContainer.handler_class(@type).new(@attributes, @bot)
-      return [nil, nil] unless dummy_handler.matches?(event)
+      return [nil, nil] unless event.instance_of?(@type) && dummy_handler.matches?(event)
 
       should_delete = nil
       should_delete = true if (@block && @block.call(event) != false) || !@block


tripl3dogdare:
While I understand the reason for this fix, wouldn't this break event handling for `ReactionEvent` to get either (for example)? Or am I just misunderstanding?

z64:
That sounds about right. That could be fixed by raising ReactionEvent ourselves I believe.

tripl3dogdare:
So essentially something along the lines of adding an Await for all three types of ReactionEvent and raising a generic RE on either of the specific ones? If so, that almost seems like a worse hack than needing to check that it's the correct subclass... If you like, I can look into a way to resolve this that doesn't require either. My Java experience is nagging at me that there has to be a way to check accurately that the type of the event is either exactly what we asked for or a subclass of what we asked for.

z64:
Yes, I mean that we would raise ReactionEvent - so if you wanted to catch multiple kinds of RE would would await ReactionEvent, and then do a `case` on the class. Or just register individual awaits for any RE subclasses. 

I don't have any issue with this as currently reaction awaits do not work right *at all* (#422) and causes some awaits to fire multiple times (#430). Since awaits are niche API (and pretty lame), im fine with fixing these broken behaviors if it means potentially (very unlikely) complicating an even less likely piece of code in the wild, that would have a very small refactoring cost anyways.

tripl3dogdare:
I'm not sure I'm understanding you fully, could you clarify? Does this break intentional use of inheritance (i.e. awaiting `ReactionEvent` getting both `ReactionAddEvent` and `ReactionRemoveEvent` as you would expect), or does it not?

With that said, yes, I agree that breaking intentional use of inheritance is better than having broken and incorrect *un*intentional use of inheritance. However, I'd still like to take a look into finding a way to not breaking inheritance at all, since I personally have found Awaits to be quite useful in the past, if you don't mind.

z64:
I mean that, with the PR as it is *right now*, yes, you could not await `ReactionEvent` anymore as it is never specifically raised, and the await would never be called.

If we raise ReactionEvent ourselves, the following is still valid:
```rb
bot.await(:foo, Discordrb::Events::ReactionEvent) do |event|
  # Raised on every kind of reaction event (Add, Remove, RemoveAll? forget what its called)
end

bot.await(:foo, Discordrb::Events::ReactionAddEvent) do |event|
  # Raised *only* on ReactionAddEvent
end
```

tripl3dogdare:
Alright, it seems after a bit of testing that it does work as intended. My apologies, I wasn't understanding you properly. Consider me satisfied.

z64:
Well, actually, just raising ReactionEvent means it *will* get the await called on every event, but I don't think you can deduce what kind of event it was further from that. [This was the patch I was thinking of based on this PR](https://gist.github.com/z64/f7f2d9328bc3d7b9ada86c4fb8d56d51), and you get `ReactionEvent` - that's it. Hmm :smile:

z64:
Going to merge this as-is for now so that particular awaits are not inherently broken as observed. We can improve upon the inheritance at a later time; @tripl3dogdare if you could look into that, especially if you already have code that relies on inheritance in awaits, that would be great.

tripl3dogdare:
Last time I checked on it everything worked as intended, may be some edge cases to check on though. I'll give it a look when I have the chance.

z64:
What doesn't work at the moment is this:
```rb
bot.await(:foo, Discordrb::Events::ReactionEvent) do |event|
  # Raised on every kind of reaction event (Add, Remove, RemoveAll? forget what its called)
end
```
Since we don't raise ReactionEvent.

If we just do this (raise ReactionEvent) the above snippet would work but you would not be able to tell what *kind* of reaction event it was (add, remove)

tripl3dogdare:
Ah, that might be slightly different - I tested against raw events rather than awaits, which works properly (event listeners for ReactionEvent get all sub events). I'll double check on the awaits situation when I get the chance.

PR #432, opened 2017-11-17
Implement the new statuses

PixeLInc:
Add the new Listening & Watching statuses

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -444,12 +444,12 @@ module Discordrb
     end
 
     # Updates presence status.
-    # @param status [String] The status the bot should show up as.
-    # @param game [String, nil] The name of the game to be played/stream name on the stream.
+    # @param status [String] The status the bot should show up as. Can be `online`, `dnd`, `idle`, or `invisible`
+    # @param game [String, nil] The name of the game to be played/watched/listened to/stream name on the stream.
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
-    # @param game_type [Integer] The type of game status to display.
+    # @param game_type [Integer] The type of game status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
     # @see Gateway#send_status_update
     def update_status(status, game, url, since = 0, afk = false, game_type = 0)
       gateway_check
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -445,25 +445,25 @@ module Discordrb
 
     # Updates presence status.
     # @param status [String] The status the bot should show up as. Can be `online`, `dnd`, `idle`, or `invisible`
-    # @param game [String, nil] The name of the game to be played/watched/listened to/stream name on the stream.
+    # @param activity [String, nil] The name of the activity to be played/watched/listened to/stream name on the stream.
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
-    # @param game_type [Integer] The type of game status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
+    # @param activity_type [Integer] The type of activity status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
     # @see Gateway#send_status_update
-    def update_status(status, game, url, since = 0, afk = false, game_type = 0)
+    def update_status(status, activity, url, since = 0, afk = false, activity_type = 0)
       gateway_check
 
-      @game = game
+      @activity = activity
       @status = status
       @streamurl = url
-      type = url ? 1 : game_type
+      type = url ? 1 : activity_type
 
-      game_obj = game || url ? { 'name' => game, 'url' => url, 'type' => type } : nil
-      @gateway.send_status_update(status, since, game_obj, afk)
+      activity_obj = activity || url ? { 'name' => activity, 'url' => url, 'type' => type } : nil
+      @gateway.send_status_update(status, since, activity_obj, afk)
 
       # Update the status in the cache
-      profile.update_presence('status' => status.to_s, 'game' => game_obj)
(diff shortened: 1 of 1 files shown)


lib/discordrb/bot.rb
```diff
@@ -449,14 +449,15 @@ def parse_mention(mention, server = nil)
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
+    # @param game_type [Integer] The type of game status to display.
```
z64: I feel we should document the current valid ranges for `game_type`, as well as `status`.

I can't review it, but the docstring for `game` should be updated to include watching and listening somehow.

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
     end
 
     # Updates presence status.
-    # @param status [String] The status the bot should show up as.
-    # @param game [String, nil] The name of the game to be played/stream name on the stream.
+    # @param status [String] The status the bot should show up as. Can be `online`, `dnd`, `idle`, or `invisible`
+    # @param game [String, nil] The name of the game to be played/watched/listened to/stream name on the stream.
```
z64: `@param activity`, `The name of the activity to be..`

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
...
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
+    # @param game_type [Integer] The type of game status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
```
z64: `activity_type`, `The type of activity status to display..`

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
...
     # @see Gateway#send_status_update
-    def update_status(status, game, url, since = 0, afk = false)
+    def update_status(status, game, url, since = 0, afk = false, game_type = 0)
```
z64: `activity`, `activity_type = 0`

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
...
       gateway_check
 
       @game = game
       @status = status
       @streamurl = url
-      type = url ? 1 : 0
+      type = url ? 1 : game_type
```
z64: `1 : activity_type`

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
...
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
+    # @param game_type [Integer] The type of game status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
     # @see Gateway#send_status_update
-    def update_status(status, game, url, since = 0, afk = false)
+    def update_status(status, game, url, since = 0, afk = false, game_type = 0)
       gateway_check
 
       @game = game
```
z64: `@activity`

lib/discordrb/bot.rb
```diff
@@ -444,19 +444,20 @@ def parse_mention(mention, server = nil)
...
       @status = status
       @streamurl = url
-      type = url ? 1 : 0
+      type = url ? 1 : game_type
 
       game_obj = game || url ? { 'name' => game, 'url' => url, 'type' => type } : nil
```
z64: `activity_object`, etc

lib/discordrb/bot.rb
```diff
@@ -444,25 +444,26 @@ def parse_mention(mention, server = nil)
...
 
     # Updates presence status.
-    # @param status [String] The status the bot should show up as.
-    # @param game [String, nil] The name of the game to be played/stream name on the stream.
+    # @param status [String] The status the bot should show up as. Can be `online`, `dnd`, `idle`, or `invisible`
+    # @param activity [String, nil] The name of the activity to be played/watched/listened to/stream name on the stream.
     # @param url [String, nil] The Twitch URL to display as a stream. nil for no stream.
     # @param since [Integer] When this status was set.
     # @param afk [true, false] Whether the bot is AFK.
+    # @param activity_type [Integer] The type of activity status to display. Can be 0 (Playing), 1 (Streaming), 2 (Listening), 3 (Watching)
```
LikeLakers2: Suggestion: Allow `:playing`, `:streaming`, `:listening`, and `:watching` as respective aliases. Not a big deal to not include them, but it'd probably be a bit more readable if people could use those symbols in their code rather than numbers.

lib/discordrb/bot.rb
```diff
@@ -474,6 +475,26 @@ def game=(name)
       name
     end
 
+    alias_method :playing=, :game=
+
```
LikeLakers2: Probably wouldn't hurt to have a `#stream=`/`#streaming=` method around here, since we have methods for the other three activity types.

lib/discordrb/bot.rb
z64: This is a "lower level" method, and I don't mind leaving it as-is for now. The values are documented, but its likely most will use the more friendly methods anyways since presence API behavior is harder to grasp.

lib/discordrb/bot.rb
z64: This doesn't work; Stream presence needs name and URL to be valid. This can't be done with `=` methods. It would work if we had a stream URL cached or just set the name of the stream to the URL, but this is pretty bad UX.

z64:
Thank you!

lib/discordrb/bot.rb
LikeLakers2: I don't particularly get what would be bad UX about allowing them to pass two strings instead of one, just like they're already doing for the other three `=` methods.

Seriously though, it wouldn't be that different from the other three methods.
```rb
def streaming=(name, url)
  gateway_check
  update_status(@status, name, url, nil, nil, 1)
  name
end
```

PR #438, opened 2017-12-12
Use debug logger for import time version message.

dreid:
This message doesn't seem necessary at import time.

### DIFF
--- a/lib/discordrb/websocket.rb
+++ b/lib/discordrb/websocket.rb
@@ -2,6 +2,8 @@
 
 require 'websocket-client-simple'
 
+Discordrb::LOGGER.debug "Using WSCS version: #{WebSocket::Client::Simple::VERSION}"
+
 # The WSCS module which we're hooking
 # @see Websocket::Client::Simple::Client
 module WebSocket::Client::Simple
--- a/lib/discordrb/websocket.rb
+++ b/lib/discordrb/websocket.rb
@@ -2,8 +2,6 @@
 
 require 'websocket-client-simple'
 
-Discordrb::LOGGER.debug "Using WSCS version: #{WebSocket::Client::Simple::VERSION}"
-
 # The WSCS module which we're hooking
 # @see Websocket::Client::Simple::Client
 module WebSocket::Client::Simple
@@ -31,6 +29,8 @@ module Discordrb
     # @param error_handler [#call] The handler that should be called when an error occurs in another handler. The error
     #   will be passed as the first parameter to the handler.
     def initialize(endpoint, open_handler, message_handler, close_handler, error_handler)
+      Discordrb::LOGGER.debug "Using WSCS version: #{WebSocket::Client::Simple::VERSION}"
+      
       @open_handler = open_handler
       @message_handler = message_handler
       @close_handler = close_handler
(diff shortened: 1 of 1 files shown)


z64:
It is not necessary, but it is desirable for bug reports and debugging. I would like to keep it. If you want, you can move this to `Discordrb::LOGGER.debug` instead.

dreid:
@z64 Done.

dreid:
@z64 Using `Discordrb::LOGGER` at import time is annoying, any objection to just doing at websocket client initialization time?

z64:
It's probably fine, but I'll just note that this currently doesn't work at all as is:
```
$ bundle exec ruby bot.rb
/home/lune/git/discordrb/lib/discordrb/websocket.rb:5:in `<top (required)>': uninitialized constant Discordrb::LOGGER (NameError)
Did you mean?  Logger
        from /home/lune/git/discordrb/lib/discordrb/bot.rb:30:in `require'
        from /home/lune/git/discordrb/lib/discordrb/bot.rb:30:in `<top (required)>'
        from /home/lune/git/discordrb/lib/discordrb.rb:4:in `require'
        from /home/lune/git/discordrb/lib/discordrb.rb:4:in `<top (required)>'
        from bot.rb:1:in `require'
        from bot.rb:1:in `<main>'
```
The initialization and importing of the logger would have to be moved it seems.
I can't look into it further at the moment.

z64:
This runs now, but for some reason the line is not shown on `log_mode: debug`. Do you see it on your tests? Also, CI is failing for trailing whitespace.
```
[DEBUG : main @ 2017-12-12 18:48:15.471] WS thread created! Now waiting for confirmation that everything worked
[DEBUG : websocket @ 2017-12-12 18:48:15.473] Connecting
[DEBUG : websocket @ 2017-12-12 18:48:15.536] Gateway URL: wss://gateway.discord.gg/?encoding=json&v=6
[DEBUG : websocket @ 2017-12-12 18:48:15.597] Obtained socket
[IN : websocket @ 2017-12-12 18:48:15.672] {"t"=>nil, "s"=>nil, "op"=>10, "d"=>{"heartbeat_interval"=>41250, "_trace"=>["gateway-prd-main-rc6l"]}}
[DEBUG : websocket @ 2017-12-12 18:48:15.672] Hello!
[DEBUG : websocket @ 2017-12-12 18:48:15.673] Trace: ["gateway-prd-main-rc6l"]
# etc...
```

dreid:
I fixed the whitespace.

@z64 It looks like this class may never be used anywhere? From a cursory look at the code that's the only explanation I can come up with.

z64:
Yeah actually, the abstract websocket is used for VWS.
```
$ grep -nr 'WebSocket.new' lib/
lib/discordrb/voice/network.rb:302:      @client = Discordrb::WebSocket.new(
```
Forgot that we actually re-implement parts of WSCS in our own `Gateway`, and we don't wrap it for the base connection.

I guess the question is if we're okay with that debug line popping up with every new VWS connection. `:debug` is a "mess" by nature so I'm not particularly bothered by it, but I don't use voice either.

I'll fish for a second opinion before merging but this is OK with me for now.

meew0:
The reason I added this message is because when we switched from Faye to WSCS, people would for some reason have old versions of WSCS installed, which caused all sorts of hard-to-diagnose problems, so I added this message so it would be immediately visible which version is used. The wrong version thing doesn't happen anymore, so the message is not as important, but I'd still like to have it there in general, as @z64 said.

PR #446, opened 2018-01-10
Adds methods for modifying the command bot's allowed channel list

omnilord:
Adding spec to test the bot for observing the channels whitelist.  Also including a few functions to simplify the functionality for adding/removing channels to the whitelist.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -333,6 +333,30 @@ module Discordrb::Commands
       [@permissions[:users][user.id] || 0, determined_level].max >= level
     end
 
+    # @see CommandBot#update_channels
+    def channels=(channels)
+      update_channels(channels)
+    end
+
+    # Update the list of channels the bot accepts commands from.
+    # @param channels [Array<String, Integer, Channel>] The channels this command bot accepts commands on.
+    def update_channels(channels = [])
+      @attributes[:channels] = Array(channels)
+    end
+
+    # Add a channel to the list of channels the bot accepts commands from.
+    # @param channel [String, Integer, Channel] The channel name, integer ID, or `Channel` object to be added
+    def add_channel(channel)
+      return if @attributes[:channels].find { |c| channel.resolve_id == c.resolve_id }
+      @attributes[:channels] << channel
+    end
+
+    # Remove a channel from the list of channels the bot accepts commands from.
+    # @param channel [String, Integer, Channel] The channel name, integer ID, or `Channel` object to be removed
+    def remove_channel(channel)
+      @attributes[:channels].delete_if { |c| channel.resolve_id == c.resolve_id }
+    end
+
     private
 
     # Internal handler for MESSAGE_CREATE that is overwritten to allow for command handling
@@ -404,14 +428,9 @@ module Discordrb::Commands
     def channels?(channel, channels)
       return true if channels.nil? || channels.empty?
       channels.any? do |c|
-        if c.is_a? String
-          # Make sure to remove the "#" from channel names in case it was specified
-          c.delete('#') == channel.name
-        elsif c.is_a? Integer
-          c == channel.id
-        else
-          c == channel
-        end
+        # if c is string, make sure to remove the "#" from channel names in case it was specified
+        return true if c.is_a?(String) && c.delete('#') == channel.name
+        c.resolve_id == channel.resolve_id
       end
     end
 
--- a/spec/commands_spec.rb
+++ b/spec/commands_spec.rb
@@ -1,47 +1,305 @@
 require 'discordrb'
 
-describe Discordrb::Commands do
-  describe Discordrb::Commands::CommandBot do
-    def command_event_double
-      event = double('event')
+describe Discordrb::Commands::CommandBot, order: :defined do
+  SIMPLE_RESPONSE = 'hi'.freeze
+  TEST_CHANNELS = [
+    1,
+    2,
+    '3',
+    'channel-four',
+    '#channel-five'
+  ].freeze
+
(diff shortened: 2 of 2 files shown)


spec/commands_spec.rb
```diff
@@ -31,19 +65,101 @@ def command_event_double
...
         it 'should return the response' do
           result = bot.execute_command(:name, command_event_double, [], false, false)
 
-          expect(result).to eq RESPONSE
+          expect(result).to eq SIMPLE_RESPONSE
         end
       end
     end
+
+    context 'no channels initialized', order: :defined do
```
z64: I'm thinking all these should be wrapped in:
```rb
describe '@attributes[:channels]' do
  # ..
end
```
or similar. It looks out of place / unclear until you look closely that that's exactly what you're testing

spec/commands_spec.rb
```diff
@@ -1,13 +1,47 @@
+# rubocop:disable Style/NumericLiterals
 require 'discordrb'
 
 describe Discordrb::Commands do
   describe Discordrb::Commands::CommandBot do
+    FIRST_CHANNEL = 41771983423143937
+    SECOND_CHANNEL = 155101607195836416
+    THIRD_CHANNEL = 123456789012345678
```
z64: Perhaps you should test for each possible type of object in `@attributes[:channels]`.
- Integer ID
- Channel object (any `#resolve_id` object)
- Channel name

lib/discordrb/commands/command_bot.rb
```diff
@@ -404,14 +423,9 @@ def required_roles?(member, required)
...
-          # Make sure to remove the "#" from channel names in case it was specified
-          c.delete('#') == channel.name
-        elsif c.is_a? Integer
-          c == channel.id
-        else
-          c == channel
-        end
+        # if c is string, make sure to remove the "#" from channel names in case it was specified
+        # Otherwise resolve the ids and compare
+        c.resolve_id == channel.resolve_id || (c.is_a?(String) && c.delete('#') == channel.name)
```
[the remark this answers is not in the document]
omnilord: `channels?` is a private method.  It is actually only used twice (as far as my search could find) in a single public function, `CommandBot#execute_command`, that is the only command of interest being executed in the provided specs.  So, essentially, these specs are implicitly testing `channels?` indirectly through `CommandBot#execute_command`.

spec/commands_spec.rb
```diff
@@ -31,17 +83,192 @@ def command_event_double
...
+        bot = Discordrb::Commands::CommandBot.new token: '', help_available: false
+
+        bot.command :name do
+          SIMPLE_RESPONSE
+        end
+
+        it 'should respond on all channels when no channel list is provided' do
+          expect(bot.attributes[:channels]).to be_empty
+
+          plain_event1 = command_event_double_for_channel(FIRST_CHANNEL)
```
Daniel-Worrall: The doubles defining should be handled by using `context`s or `behaves` if they share the same kind of behaviour

spec/commands_spec.rb
```diff
@@ -31,17 +83,192 @@ def command_event_double
...
         it 'should return the response' do
           result = bot.execute_command(:name, command_event_double, [], false, false)
 
-          expect(result).to eq RESPONSE
+          expect(result).to eq SIMPLE_RESPONSE
+        end
+      end
+    end
+
+    describe '@attributes[:channels]' do
```
Daniel-Worrall: Even though the methods being tested are related to this, we prefer to group our tests by file->class->method so something like `describe 'Channel'` or `describe '#name='`

z64: Yeah, thinking about this again, it should probably `describe '#execute_command'` with subsequent `context`.

spec/commands_spec.rb
```diff
@@ -223,49 +253,48 @@ def command_event_double_with_channel(channel)
...
-
-          event = command_event_double_for_channel(FIRST_CHANNEL)
-          result = bot.execute_command(:name, event, [])
-          expect(result).to eq SIMPLE_RESPONSE
-
-          event = command_event_double_for_channel(THIRD_CHANNEL)
-          result = bot.execute_command(:name, event, [])
-          expect(result).to be_nil
-
+        it 'new channels should replace old channels' do
```
LikeLakers2: `it 'should replace old channels with new channels' do`?

spec/commands_spec.rb
```diff
@@ -1,148 +1,151 @@
...
+    context 'with channel filter', order: :defined do
+      context 'when list is not initialized in bot parameters', order: :defined do
         bot = Discordrb::Commands::CommandBot.new token: '', help_available: false
 
         bot.command :name do
           SIMPLE_RESPONSE
         end
 
-        it 'should respond on all channels when no channel list is provided' do
+        it 'has no channels' do
```
LikeLakers2: I don't think we should make these statements definitive -- it makes more sense to say "it should" in these contexts rather than "it does" or "it does not", because we're testing functionality (making sure it works), not trying to prove a point to some committee.

This comment applies to all the specs here, not just this one.

omnilord: Okay, however, the spec you linked me too says otherwise: http://www.betterspecs.org/#should

LikeLakers2: Ah, nevermind then. I was basing that comment off what I had seen in the specs before.

spec/data/channel_filter_list.json
```diff
@@ -0,0 +1,14 @@
+{
```
z64: I think this data file is kind of arbitrary. I would like to keep the `data` JSON files to fake pieces of "actual" data from discord / the web.

Could we just construct this inside the ruby spec instead?

PR #459, opened 2018-03-14
raise if embed limit over

soukouki:
https://discordapp.com/developers/docs/resources/channel#embed-limits

### DIFF
--- a/lib/discordrb/webhooks/embeds.rb
+++ b/lib/discordrb/webhooks/embeds.rb
@@ -17,7 +17,7 @@ module Discordrb::Webhooks
       @author = author
       self.fields = fields
     end
-    
+
     # see https://discordapp.com/developers/docs/resources/channel#embed-limits
     TITLE_LIMIT = 256
     DESCRIPTION_LIMIT = 2048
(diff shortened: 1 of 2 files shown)


Daniel-Worrall:
This is already being covered in #338. Note that this PR is not abandoned, it just needs a lot of work done to it so it is slow coming and there have been more pressing things come up.

I prefer 338 over this implementation because this only checks the rates on the embed builder when you can send embeds through many different ways, and also cannot be turned off (Which if embed limits change, users can bypass the warning pretty easily instead of having to patch in the source or wait for an update. Especially if they're using gem and not master.)

As an aside, the implementation of this isn't very Rubyesque.

Anyway, if you'd like to know what we're doing up to date, you should join us on the [discord channel](https://discord.gg/cyK3Hjm) where we're discussing more than is displayed on the Issue Tracker.

z64:
Yeah I'm with @Daniel-Worrall on this one.

I seriously appreciate your interest in contributing to this project, but as Daniel explained I don't think this implementation is going to work for us.

Not everyone uses these methods on the builder objects, so anyone (like myself in my own bots) who doesn't use them loses out on the format checking. We shouldn't force users to use the embed builder. This can be done by having a single point of checking, when the `Embed` is serialized into a hash before being passed to the API, and that's exactly the approach I took in #338.

This is definitely something I want, and knowing someone else is interested in it I will try to wrap up #338 sometime soon.

So thank you again, and please do consider joining us in the Discord channel if you'd like to discuss helping us out with some other things, as we'd love more contributors. :smile:

PR #460, opened 2018-03-16
Document `discordrb-webhooks` in README

z64:


README.md
```diff
@@ -126,6 +127,33 @@ You can find examples of projects that use discordrb by [searching for the disco
 
 If you've made an open source project on GitHub that uses discordrb, consider adding the `discordrb` topic to your repo!
 
+## Webhooks Client
+
+Also included is a webhooks client, which can be used as a separate gem `discordrb-webhooks`. This special client can be used to form request to Discord webhook URLs in a high-level manner.
```
VxJasonxV: > requests

VxJasonxV:
Is it worth linking to https://leovoel.github.io/embed-visualizer/ in here too? If not here then somewhere involving webhooks? It's an incredible resource.

z64:
I suppose, but that isn't only limited to webhooks. It applies to bots as well, so :thinking: (`#send_message`, `#send_embed`, etc.). Feels more like a wiki thing to me. If we can't think of a better place I'm not opposed to adding it there.

PR #469, opened 2018-03-24
Alternative for restricting commands to some roles instead of all of required_roles

elfenars:
The current logic of the `required_roles?` restricts the command to be executed only when the user has **all** of the roles specified in the attribute.

This not only is non consistent with the [Wiki](https://github.com/meew0/discordrb/wiki/Events#attributes), but also doesn't allow much flexibility when allowing more than one role to execute the command.

i.e:

```
command :blah, required_roles: [123, 456] do |e|
  ...
end
```

The current logic would require the user to have both roles 123 **and** 456, when ideally it should be that the user has **either** 123 or 456.

This would allow for users of both roles to execute the comand.

I think this is more common usage when defining commands than "I just want this command to be ran by people with multiple roles".

Since this is the most common case I believe, if someone _really_ wants to allow users of exactly [123, 456] to run the command, a simple check can be done inside the command definition.

Another option would be to re-define the method params to allow behavior switching depending of the case, but I believe this to be overhead and not necessary.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -415,9 +415,9 @@ module Discordrb::Commands
     end
 
     def required_roles?(member, required)
-      return (required.nil? || required.empty?) if member.webhook? || member.is_a?(Discordrb::Recipient)
+      return true if member.webhook? || member.is_a?(Discordrb::Recipient) || required.nil? || required.empty?
       if required.is_a? Array
-        required.all? do |role|
+        required.any? do |role|
           member.role?(role)
         end
       else
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -156,7 +156,7 @@ module Discordrb::Commands
           result
         else
           available_commands = @commands.values.reject do |c|
-            !c.attributes[:help_available] || !required_roles?(event.user, c.attributes[:required_roles]) || !required_permissions?(event.user, c.attributes[:required_permissions], event.channel)
+            !c.attributes[:help_available] || !required_roles?(event.user, c.attributes[:required_roles]) || !allowed_roles?(event.user, c.attributes[:allowed_roles]) || !required_permissions?(event.user, c.attributes[:required_permissions], event.channel)
           end
           case available_commands.length
           when 0..5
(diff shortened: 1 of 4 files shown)


LikeLakers2:
I disagree with this change. While I *understand* it, this is a breaking change, too. Perhaps we could include a separate flag to say "needs any required_role" or "needs all required_roles", with the default being needing all required_roles like it's always been.

elfenars:
This is what I meant by 

> Another option would be to re-define the method params to allow behavior switching depending of the case, but I believe this to be overhead and not necessary.

You're right, this would break logic of previous devs who already took this logic for granted so maybe the flag is needed. I could provide an alternative solution if this is the preferred way to go.

LikeLakers2:
Ah, I missed that bit of your comment. My bad. Yeah, I would prefer having a switch for it to avoid breaking current code. That seems like the best option right now.

elfenars:
This is a proposal for what you say.

z64:
> Since this is the most common case I believe, if someone really wants to allow users of exactly [123, 456] to run the command, a simple check can be done inside the command definition.

The same could be said for the behavior added in this PR.
```rb
bot.command(:foo) do |event|
  next unless event.user.roles.any? { |r| [123, 456].include? r.id }
  # do whitelisted behavior
end
```
I'm 50/50 on adding more options like this. We've talked about rewriting the command bot in the future (I don't know the details on this), and I'm really interested in restructuring entirely how we handle arbitrary checks like this so that it is more easily extendable by users.

/ramble

Anyways, we don't have tests for this behavior (`required_roles`, `allowed_roles`), they should be added to track if this breaks so we can refactor or remove it later.

elfenars:
I tried, but the commands_spec is honestly very spaghetti-ish for me to understand right now. There's too many doubles and dependencies to one another and I can't completely wrap my head around it to make the tests myself.

Some guidance or examples would be pretty nice.

z64:
Sure. You can drop me a line on discord (z64#1337) and maybe we can set you in the right direction.

elfenars:
Ok, this is hacky af, but that's as close as I got to get it working.


1. I added 2 roles, loaded them from the role.json file and then had to change the ID of one of them so I don't have the same ID on both.
2. I added roles to the author double and permited the `role?` method to be called on it. (Here's where the problem lies)
3. I added the test cases.

I "could" just stub the `roles?` method, but that would defeat the purpose and it wouldn't allow me to correctly test if the user has the roles or not (and if the logic of `required_roles` or `allowed_roles` is correct).

I can't call the original method either because the author is a pure double and not a partial double, so Rspec doesn't allow me to pass `.and_call_original` to the double.

_Technically_ if we call the original method, the tests should pass, if we don't call the original method, then I would need to replicate its behaviour in the double which is stupid.

I hope this is enough info.

z64:
Sorry for stalling on this! :disappointed:  I'm still fine with this being added; I'll take care of any additional changes needed to merge this - I think we might just forgo the specs unless @Daniel-Worrall has any bright ideas. Thank you for your patience!

Issue #455, opened 2018-02-02
Ruby 2.5 Support on Windows


z64:
Currently, discordrb appears to fail to install on Windows because the following gems are not yet available under Ruby 2.5:
- `ffi`
```
ffi requires Ruby version < 2.5, >= 2.0. The current ruby version is 2.5.0
```
- `unf_ext`
```
unf_ext requires Ruby version < 2.5, >= 2.0. The current ruby version is 2.5.0
```
Ruby 2.4 can be used until these are updated or we find some kind of workaround.

(To be clear, Linux seems to be installing fine)

Daniel-Worrall:
After going over this issue, it appears only the gem version uses `unf_ext`. Either a remnant of an old yard version or something else. I wasn't able to locate exactly where it's pulling it from as I'm not able to reproduce the error through `bundle install`, only through `gem install`.

As for master, it is still plagued by `ffi` being for previous Ruby versions. This is fixed in the master version of `ffi` but is not yet released as a gem version, unfortunately.

I went through the effort of locating each gem that uses ffi and had them forked, then patched in for them to require from master. I'm not convinced that it won't be broken when running a bot through it but it will at least compile.

If someone would like to take advantage on this (Keeping in mind, this is only a problem on Windows Ruby 2.5 and you can just use 2.4 or below instead), you can use this fork in bundler:
```gem 'discordrb', git: 'https://github.com/y32/discordrb.git', branch: '2.5'```

z64:
We've upgraded our RestClient dependency in #478 and this has enabled us to run discordrb on Windows, but we're still testing the stability of it. If any of our master branch users experience uncommon issues on 2.5 in Windows using these updated dependencies, please feel free to report them here or in our chat.

z64:
We haven't had any reports of issues since upgrading. I think we can close this for now, if anything new comes up, an issue can be opened up with more specific details.

PR #474, opened 2018-03-31
Adding RestClient::Utils patch for Ruby versions >=2.5 on Windows

swarley:
This fix has been tested to work on Windows 7/8/10 with Ruby versions 2.5.0 and 2.5.1

Ref #455

### DIFF
--- a/lib/rest-client/utils.rb
+++ b/lib/rest-client/utils.rb
@@ -1,31 +1,30 @@
-# @note In Ruby versions greater than 2.4 on Windows, Discordrb's 
-# environment does not play nicely with the use of enumerators in
-# this specific function. The exact reason is unknown, but changing the use of
-# `Enumerator#next` into iteration via `Array#shift` prevents a silent crash.
+# RestClient module, RestClient#get being the main scope of the issue
+module RestClient
+  # Utils contains the helper method that causes issues
+  module Utils
+    # @note In Ruby versions greater than 2.4 on Windows, Discordrb's 
+    # environment does not play nicely with the use of enumerators in
+    # this specific function. The exact reason is unknown, but changing the use of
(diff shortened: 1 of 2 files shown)


lib/rest-client/utils.rb
```diff
@@ -0,0 +1,33 @@
+if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.5') && ENV['OS'] == 'Windows_NT'
```
z64: Actually, I have a better idea. Could we move this to `lib/discordrb.rb` so that it wraps the `require`, and move your first note around it? So that it looks like this:
```rb
# lib/discordrb.rb
require 'discordrb/bot'
require 'discordrb/commands/command_bot'
require 'discordrb/logger'

# This fix in particular addresses an issue within RestClient, confirmed
# in RestClient 2.0.2 but only within the Discordrb environment.
if Gem::Verision.new(RUBY_VERSION) >= Gem::Version.new('2.5') && ENV['OS'] == 'Windows_NT'
  require 'rest-client/utils'
end
```
(note that I removed the `@note` tag)

I think that will be more "visible".

lib/rest-client/utils.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  # @note This fix in particular addresses an issue within RestClient, confirmed
+  # in RestClient 2.0.2 but only within the Discordrb environment.
+  module RestClient
+    # @note In Ruby versions greater than 2.4 on Windows, Discordrb's
```
z64: Could you move this docstring onto the patched method itself?

z64:
As discussed in chat, I think we're all comfortable to apply #478 over this monkey patch. Thanks again so much for the effort you put in to debugging this for our windows users! :heart:

PR #476, opened 2018-04-02
[CI] Test against Ruby 2.3, 2.4 and 2.5

nicolasleger:


ChallahuAkbar:
Since you opened this PR literally an hour before I was about to, I figured I'd at least contribute and say that I'm pretty sure that we don't need that install bundler line anymore, and if we do, it should be using the latest version as well.

ChallahuAkbar:
Just tested, and [yes](https://travis-ci.org/ChallahuAkbar/discordrb/jobs/361968250#L462), Travis uses the latest bundler already.

nicolasleger:
Thanks, PR updated.

z64:
Thanks!

PR #478, opened 2018-04-04
Add rest-client dependency restriction for Windows Ruby 2.5

Daniel-Worrall:
This should fix our 2.5 woes we've been having.

See #455 #474 

Starting up a bot for 24hr testing to test if stable.

### DIFF
--- a/discordrb-webhooks.gemspec
+++ b/discordrb-webhooks.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.add_dependency 'rest-client'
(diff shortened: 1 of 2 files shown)


connorshea:
If this is a Windows-only issue, is it acceptable to use v2.0.2 on macOS/Linux? It'd be nice to limit the 2.1.0.rc1 lock to Windows if that's the case, and let 2.0.2 be used for other platforms.

z64:
That would be fine. While investigating this issue, I [looked into doing that](https://en.wikibooks.org/wiki/Ruby_Programming/RubyGems#How_to_install_different_versions_of_gems_depending_on_which_version_of_ruby_the_installee_is_using) after we found that using the release client worked. Writing an extension seemed more trouble than it was worth at the time, when having tested the release client myself on linux without any issues I didn't think about it again.

I also don't use Windows, so someone else would have to develop this extension and test that it installs without issue.

PR #479, opened 2018-04-04
Fix issue with ban reasons being mangled or rejected

tripl3dogdare:
Ban reasons, unusually, are sent in the query params rather than the request headers. This means they need to be URL encoded to work reliably. Currently this must be done manually, which this commit fixes.

### DIFF
--- a/lib/discordrb/api/server.rb
+++ b/lib/discordrb/api/server.rb
@@ -1,3 +1,5 @@
+require 'uri'
+
 # API calls for Server
 module Discordrb::API::Server
   module_function
@@ -187,7 +189,7 @@ module Discordrb::API::Server
       :guilds_sid_bans_uid,
       server_id,
       :put,
-      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}?delete-message-days=#{message_days}#{reason ? "&reason=#{reason}" : ''}",
+      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}?delete-message-days=#{message_days}#{reason ? "&reason=#{URI.escape(reason)}" : ''}",
       nil,
       Authorization: token
     )
--- a/lib/discordrb/api/server.rb
+++ b/lib/discordrb/api/server.rb
@@ -1,4 +1,4 @@
-# API calls for Server
+e# API calls for Server
 module Discordrb::API::Server
   module_function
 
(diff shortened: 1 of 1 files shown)


omnilord:
I could be wrong about this, but there is probably a more pressing question of why this seems to be the only endpoint that encodes data for a mutable operation in the query string.  Every other API endpoint for PUT, POST, and PATCH use parameter after the URL ([nil here](https://github.com/meew0/discordrb/pull/479/files#diff-8cf684da3bdab9b2b2ff5ed6753bef8dR193)) to encode a JSON string in the body of the request.

tripl3dogdare:
I've applied the requested changes. Let me know if there's anything else you'd like me to do!

lib/discordrb/api/server.rb
```diff
@@ -1,4 +1,4 @@
-# API calls for Server
+e# API calls for Server
```
Daniel-Worrall: Pls.

lib/discordrb/api/server.rb
```diff
@@ -183,11 +183,12 @@ def bans(token, server_id)
   # Ban a user from a server and delete their messages from the last message_days days
   # https://discordapp.com/developers/docs/resources/guild#create-guild-ban
   def ban_user(token, server_id, user_id, message_days, reason = nil)
+    reason = URI.encode(reason) if reason
     Discordrb::API.request(
       :guilds_sid_bans_uid,
       server_id,
       :put,
-      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}?delete-message-days=#{message_days}#{reason ? "&reason=#{reason}" : ''}",
+      "#{Discordrb::API.api_base}/guilds/#{server_id}/bans/#{user_id}?delete-message-days=#{message_days}#{reason}" : ''}",
```
omnilord: 1. Syntax error at `:`.  I have found that running `bundle exec rspec` and  `bundle exec rubocop` before commits helps find defects.
2. You are now missing `&reason=` from the query string.

lib/discordrb/api/server.rb
omnilord: ...

tripl3dogdare: My apologies, my changes were made rather hastily and through the GitHub website rather than a local copy. I'll clean it up a bit when I get home.

tripl3dogdare:
Sorry this commit took so long, I completely forgot about this PR. It should be cleaned up properly now, let me know if there's anything else you want me to change.

PR #486, opened 2018-04-22
Added `details` field to Events::PlayingEvent

xTVaser:
When working with Twitch typed `PlayingEvents` there was no way to get the name of the game that was being streamed.  This information is contained within the optional `details` field https://discordapp.com/developers/docs/topics/gateway#activity-object-activity-structure and was just unused.

### Changes
Added `details` to event object.
Noticed that the `url` field from the API is not guaranteed to be included on the `game` object so added an extra check there.

### DIFF
--- a/lib/discordrb/events/presence.rb
+++ b/lib/discordrb/events/presence.rb
@@ -65,17 +65,22 @@ module Discordrb::Events
     # @return [String] the URL to the stream
     attr_reader :url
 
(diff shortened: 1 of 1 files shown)


z64:
I'm fine with this, but it's worth noting this implementation needs to be cleaned up at some point. The Activity stuff is relatively new in the API and this isn't the cleanest way of supporting it (we shouldn't flatten the game object like that..). A review of the events system is on our roadmap anyways.

PR #487, opened 2018-04-28
Appease the RuboCop

ChallahuAkbar:
Not quite finished with this but I updated the version of RuboCop, cleaned up the config file, and applied the new cops. There are a few left in `.rubocop_todo.yml` still that I wasn't totally sure what I should do about. Otherwise, I did my best to make sure I wasn't breaking anything but it definitely needs looking at by someone more familiar with the codebase. Also, I tried to avoid "editorial" decisions, so as much as possible I just applied whatever the RuboCop defaults were, meaning there may be some style decisions that need to be made.


.rubocop.yml
```diff
@@ -1,65 +1,51 @@
...
   Enabled: false
 
-# Allow CamelCase constants as well as SCREAMING_SNAKE_CASE
-Style/ConstantName:
-  Enabled: false
-
-# Allow 'Pokémon-style' exception handling
-Lint/RescueException:
-  Enabled: false
+# So RuboCop doesn't complain about application IDs
```
z64: I wouldn't be opposed to just disabling this anywhere. The only long numeric literals we have are snowflakes, and place value makes zero sense on them since they're just an opaque value with embedded binary data.

.rubocop_todo.yml
```diff
@@ -0,0 +1,33 @@
+# This configuration was generated by
```
z64: > Note that changes in th einspected code, or installation of new versions of RuboCop may require this file to be generate again

I don't see the point of checking this in to VCS. I'm not going to expect every PR to come with an update to this, and I don't think its practical to do with CI. We should just decide whether to disable these, or fix them. Unless I'm misunderstanding something, which is entirely likely at this hour.

examples/eval.rb
```diff
@@ -12,7 +12,7 @@
 
   begin
     eval code.join(' ')
-  rescue
+  rescue StandardError
```
z64: I don't know how to put the audible "ehhhh..." noise I made into a productive comment.

For the library I don't care, but for the eval command I'm not sure about this.

lib/discordrb/permissions.rb
```diff
@@ -5,7 +5,7 @@ module Discordrb
   class Permissions
     # This hash maps bit positions to logical permissions.
     # I'm not sure what the unlabeled bits are reserved for.
-    Flags = {
+    FLAGS = {
```
z64: Just for my own notes, this is a breaking change. I don't feel very strongly about it either way.

spec/mock/api_mock.rb
```diff
@@ -18,7 +18,7 @@ def raw_request(type, attributes)
                    end
       @last_headers = attributes.last
 
-      @next_response ||= nil
+      @next_response ||= nil # rubocop:disable Naming/MemoizedInstanceVariableName
```
z64: Isn't this (whole) line pointless?

.rubocop_todo.yml
ChallahuAkbar: Yeah, I agree, the reason I kept it in the PR for now is that basically it's not done, as I was holding off til I could get feedback before acting on those last 3 cops in the todo file.

spec/mock/api_mock.rb
ChallahuAkbar: ¯\\\_(ツ)_/¯ went ahead and removed it.

examples/eval.rb
ChallahuAkbar: Yeah, this is one of the ones I thought was kinda iffy. I ended up just going with the default, and explicit rescue also had fewer changes project-wide. Generally though I don't really have an opinion on this other than I think it should be consistent.

.rubocop.yml
ChallahuAkbar: Yeah, fair, I can't think of anything else that might be added, but since the only place it's currently triggering is the examples, I think it might be better to just leave it at excluding the examples and if we end up with literals we want to ignore in the actual lib code, then we can disable it globally, but it also might turn out that we end up with some other large literal that we actually would like this to trigger on. Basically, I'll leave it up to you but I don't think there's any harm in just leaving it the way it is.

.rubocop_todo.yml
```diff
@@ -0,0 +1,33 @@
...
+# `rubocop --auto-gen-config`
+# on 2018-04-28 05:24:29 -0600 using RuboCop version 0.55.0.
+# The point is for the user to remove these configuration records
+# one by one as the offenses are removed from the code base.
+# Note that changes in the inspected code, or installation of new
+# versions of RuboCop, may require this file to be generated again.
+
+# Offense count: 1
```
ChallahuAkbar: So I wasn't totally sure on this one. It's getting tripped up by the fact that the rubocop config file sets `TargetRubyVersion: 2.2` which is great for the base gem; however, the webhooks gemspec specifies a minimum version of `2.1.0`. Should I update the minimum version in the webhooks gemspec, or should this issue be ignored?

.rubocop_todo.yml
```diff
@@ -0,0 +1,33 @@
...
+# Configuration parameters: Include.
+# Include: **/*.gemspec
+Gemspec/RequiredRubyVersion:
+  Exclude:
+    - 'discordrb-webhooks.gemspec'
+
+# Offense count: 3
```
ChallahuAkbar: Honestly, I left this one out because I'm just confused about the warnings. From what I can tell, rubocop is flagging these methods because the only way to get the return value of a setter is with a call to `super`. So, given that, what should be done here?

z64:
Just as a note (mostly to my future self): As discussed in chat, we'll merge this after 3.3 is released, before #482

ChallahuAkbar:
Rebased and updated!

z64:
I think we need to add `Style::FrozenStringLiteralComment` to this.

ChallahuAkbar:
Updated and rebased.

ChallahuAkbar:
Redated and upbased!

z64:
Gave a once-over on my phone, looks good! Will take one last pass when I get home then I think this is G2G 👍

z64:
Thanks for the PR & your patience when updating everything! 🙇

PR #489, opened 2018-04-30
Update to new Code Climate test reporter

ChallahuAkbar:
The Code Climate test reporter gem that currently being used is deprecated; this PR changes the CI config to use the [new method](https://docs.codeclimate.com/docs/configuring-test-coverage).

Please note that when these changes are made, the env variable set in Travis will need to change to `CC_TEST_REPORTER_ID`; see the above link for more details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,12 @@ rvm:
   - 2.3.7
   - 2.4.4
   - 2.5.1
+before_script:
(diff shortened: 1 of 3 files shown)


ChallahuAkbar:
Was discussed in discord, but figured I should note here that this requires @meew0 to get the repo token from Code Climate, since no one else can.

PR #495, opened 2018-05-16
Introduce Master Docs GitPages

Daniel-Worrall:
See https://daniel-worrall.github.io/discordrb/ for an example

Because RubyDocs is slow as fk.


z64:
This should be done through CI.

PR #497, opened 2018-05-16
Add Channel#sort_after

z64:


spec/data_spec.rb
```diff
@@ -314,6 +318,75 @@ module Discordrb
...
+        all_channels << excluded
+        allow(server).to receive(:channels).and_return(all_channels)
+
+        expect(API::Server).to receive(:update_channel_positions)
+          .with(any_args, an_array_excluding(*[excluded].map { |e| { id: e.id, position: instance_of(Integer) } }))
+        channel.sort_after(expected_channels.last)
+      end
+
+      context 'when other is not on this server' do
+        it 'should raise ArgumentError'
```
z64: Good catch!

spec/data_spec.rb
```diff
@@ -314,6 +318,75 @@ module Discordrb
...
+      end
+
+      context 'when other is not of Channel, #resolve_id, nil' do
+        it 'should raise TypeError'
```
z64: I think everywhere else with a similar condition we are currently using ArgumentError where we should be probably be using this instead.

spec/data_spec.rb
```diff
@@ -314,6 +318,75 @@ module Discordrb
...
+      end
+
+      context 'when position doesn\'t change' do
```
z64: Nope on these - we always have to make an API call when the user asks.

spec/data_spec.rb
```diff
@@ -314,6 +318,75 @@ module Discordrb
...
+      context 'when other channel is not the same type' do
+        it 'should raise ArgumentError'
+      end
+
+      context 'when channel is not a category' do
+        context 'and when changing category' do
+          it 'should send new parent_id'
+
+          context 'with lock_permissions as false and permissions different' do
+            it 'should log that the permissions were not synced'
```
z64: Don't quite see the point in this. This is surely `debug` logging, and I can't imagine a person who is going to be looking for auditing like this when there are much more effective means.

spec/data_spec.rb
```diff
@@ -6,9 +6,13 @@
 module Discordrb
   describe Channel do
     let(:data) { load_data_file(:text_channel) }
+    # Instantiate the doubles here so we can apply mocks in the specs
```
z64: Don't think this comment is needed.

z64:
@Daniel-Worrall I'm fine with this as it is now. Thanks for getting the specs started. I'll be merging this Saturday so if you want to review the specs I filled out you can do so before then, or send a follow-up patch afterwards.

For just small changes you can of course just push to our branch in the interest of time. No need to comment/review them, just do so in their own commits so I can review for my own notes before merging.

z64:
Thanks!

PR #498, opened 2018-05-17
Add MiddlewareBot

z64:
To any watchers; please :+1: or :-1: if you like or dislike this - it will help us gage whether something like this should be merged :smiley_cat: 

---

This is a design pattern I've been using a lot lately in Crystal, and I think it could serve useful to a particular audience of discordrb "power users".

**The Problem:** `Bot` and `CommandBot` offer a limited set of options for configuring behavior and filtering incoming events. For example:
- In `bot.message`, you can use `in:`, `start_with:`.
- In `bot.command` you can use `required_roles:`, `permission_level:`

**If you want to change their behavior or add new options**, you have to:
- Write a bunch of duplicated code (`next unless...; next unless...`)
- Hack the library
- Come up with an entirely custom system that wraps `Bot`, or that your handlers dispatch to. Less experienced developers who want more out of `Bot` and `CommandBot` will find this the most difficult, and probably lead to very messy abstractions.
- Collaboration on advanced bots is more difficult because of these overly complex helpers that each developer creates.
- Your hacks that you add or implementations you create will probably be difficult to spec.

**This PR adds a new `Bot`, `MiddlewareBot`, that attempts to address all of those:**
- Instead of hacking the library, `MiddlewareBot`s handlers give you a place to inject your own code that receives a callback in a DRY way that you can re-use everywhere.
- Instead of designing a totally custom helper system from scratch, this gives you footing - a template to start with, that follows some basic rules: 
1. A `class` that responds to `def call(event, state)`. 
2. You chain these classes onto event handlers and they are executed in sequence. 
3. Whether or not your `call` method `yield`s determines if the rest of the chain is executed.
- Developers who work together using `MiddlewareBot` are now all using the same system to inject whatever custom code they want.
- Speccing middleware is *easy*: Throw whatever doubles you want down `call`.
- You can still use the regular event attributes (`in:`, `start_with:`) alongside middleware; porting an existing `Bot` is free.

---

Please read the examples and documentation I've written, as this is a simple concept that has a *lot* of implications - use your imagination when looking at these examples! If you have any questions, please review or ask me on Discord.

This could even simply replace `Bot`/`EventContainer` entirely as it would not be a breaking change at all because of the way this works!; however I figure, if we want this, it would be best as its own `Bot` class for at least one major release.

I could have written a `Container` module for this bot, but that can come in a later step.

Long term, this could also be used to clean up `Events` code internally in a really nice way, providing an easy vehicle to spec them as well.

### DIFF
--- a/examples/middleware_bot.rb
+++ b/examples/middleware_bot.rb
@@ -1,0 +1,95 @@
+require 'discordrb'
+require 'discordrb/middleware/middleware_bot'
+
+bot = Discordrb::Middleware::MiddlewareBot.new token: 'B0T.T0KEN.here'
+
+# Middleware that only yields on specific channels.
+# It can be configured to take channels by ID or by name per-handler.
+class ChannelFilter
+  def initialize(channels)
+    @channels = Array(channels)
+  end
+
+  def call(event, _state)
+    channel = event.channel
+
+    matches = @channels.any? do |c|
+      if c.is_a?(String)
+        c.delete('#') == channel.name
+      elsif c.is_a?(Integer)
+        c.id == channel.id
+      end
+    end
+
+    yield if matches
+  end
+end
+
+# Middleware that parses incoming messages and stores the results
+# in a state hash.
+class Parser
+  def initialize(prefix)
+    @prefix = prefix
+  end
+
+  def call(event, state)
+    split = event.message.content.split(' ')
+    command = split[0]
+    state[:command] = split[0][1..-1]
+    state[:arguments] = split[1..-1]
+
+    yield if command.start_with?(@prefix)
+  end
+end
+
+# Apply a middleware to an event by passing it as an argument.
+# This generates:
+#   - !add 2 3 1 (responds 6)
+#   - !ping (responds 'pong')
+bot.message(Parser.new('!')) do |event, state|
+  # Tap into the state hash set by your middleware:
+  case state[:command]
+  when 'add'
+    event.respond state[:arguments].map { |i| Integer(i) }.sum
+  when 'ping'
+    event.respond 'pong'
+  end
+end
+
+# Chain multiple middleware to create things like filters on certain
+# conditions. The following will only work in a channel named "general",
+# when the message starts with "?".
+# This generates:
+#   - ?info (responds 'made with discordrb')
+#   - ?emoji (posts a random custom emoji)
+bot.message(ChannelFilter.new('#general'), Parser.new('?')) do |event, state|
+  case state[:command]
+  when 'info'
+    event.respond 'made with discordrb'
+  when 'emoji'
+    event.respond bot.emoji.sample.to_s
+  end
+end
+
+# Fully compatible with exisitng event handler attributes. You specify them
+# after your middleware chain, just note that they will be considered *first*,
+# before your middleware is run.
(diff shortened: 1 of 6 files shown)


LikeLakers2:
So wait, I'm not sure I understand. Does this instantiate a class every time the related event happens, letting it do its thing as effectively its own little mini-session, then erasing the instance once its done?

Daniel-Worrall:
I'm liking the idea of this after a long discussion with z64 on Discord. Some things I disagree with like having only one class of Middleware being accessible to the state yield, but this could be resolved by having more complicated Middleware, handling more state in the Middleware class or using a combination Middleware that gives scope to the ones beneath it.

This is to go onto the 4.0 milestone though, and I was in disagreement that it should be pushed while we still had a HandlerMiddleware, but since this is all technically supporting old formatting, I'm okay with the go ahead.

I'm in favour too of adding CommandBot related Middlewares into the library at a later date. Also, going over all the Events we have and ensuring they have the relevant stock middlewares for the context as there are some events that are lacking important matchers.

z64:
I've updated this with some improvements to make adoption much smoother.

- MiddlewareBot is removed. After reviewing this again, `MiddlewareBot` is essentially just an `EventContainer`, so we now just use `Middleware` as a mixin.

- As a consequence of this: this PR now supports the existing container pattern! 🎉 

- All internal classes moved to `middleware.rb`, and `middleware/` will be kept for `Stock` implementations.


```rb
require "discordrb"
require "discordrb/middleware"

# Enable middleware functionality on `Bot`
# (This is optional if you are *only* using containers!)
Discordrb::Bot.include(Discordrb::Middleware)

module MyHandlers
  extend Discordrb::Middleware

  filter = lambda do |event, ctx, &block|
    block.call if event.content == "!ping"
  end

  message(filter, from: "z64") do |event|
    event.respond "pong"
  end
end

bot = Discordrb::Bot.new(token: "token")
bot.include!(MyHandlers)
bot.run
```

I also have no idea why rubocop is failing. If anyone has any ideas, please comment. I've tried a few things before I didn't have time to mess with it any longer.

z64:
The next change (665efb6) replaces the placeholder `lambda` stock middleware with a thin `MessageFilter` class that implements each of the attribute matchers as individual methods. `call` determines which matcher to use as configured by the `MessageFilter` constructor. This is effectively an optimization in the final composed `Stack` chain (`yield` > `block.call`), but it also makes the code easier to read over the noisy `lambdas`. This should be a reference for all other stdlib stock middlewares going forward.

z64:
I've squashed the development history on this now that I believe this has reached a solid point enough to be merged as-is, allowing for further improvements to come from myself / the community in later PRs in between the 3.3..4.0 cycle 🙂 

If anyone else would like clarification on anything / review (@Daniel-Worrall, a recommended way to DRY the `MessageFilter` specs would be appreciated), now would be a good time 👍 

I've rendered the documentation here if someone would like to view the docs macros, etc, as-rendered:

https://zsixtyfour.com/discordrb-middleware-docs/Discordrb/Middleware.html

ChallahuAkbar:
Just want to quickly mention that #487 is breaking and so it (and by extension this) will be merged in >= v4.0. Obviously this means that "post-3.3" is accurate, but I think it's helpful to note that this depends on upcoming breaking changes.

PR #499, opened 2018-05-18
Add blocking await methods

z64:
With this PR, we can now do this:
```rb
require "discordrb"

bot = Discordrb::Bot.new token: ENV["TOKEN"]

bot.message(with_text: '!test') do |event|
  event.respond 'What is your name?'

  # Execution of this event pauses here until a matching event is raised:
  response = event.user.await!(:"quiz_#{event.user.id}")

  event.respond "Hello #{response.content}!"
end

bot.run
```

Motivation: Lots of users look for awaits (or just ask in general) in order to provide conversational behavior, similar to how one would make a basic CLI application with `gets`. Ultimately they get confused with how awaits work and their async nature, how to persist/kill the await, or the fact that you can't do "nested" or "chained" awaits (unless you provide some of your own infrastructure) at the moment.

I think `key` is still confusing too to some users. Maybe we can intelligently provide a default or store awaits in some other way, but I won't cover that in this PR.

Currently the API is `#await!` ("bang for blocking"). I like this semantic but I'm open to other suggestions. A flag on `#await` I think isn't acceptable, this behavior is too different and needs to be distinguished. Maybe `#await_sync`, if that isn't too confusing.. perhaps an entirely different name.

I barely touched the documentation, so please suggest improvements. I think at the least `@example` should be added to `User#await`, `Channel#await`, and `Message#await`.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -559,6 +559,43 @@ module Discordrb
       @awaits[key] = await
     end
 
+    # Awaits an event, blocking the current thread until a response is received.
+    # @param type [Class] The event class that should be listened for.
+    # @option attributes [Numeric] :timeout the amount of time to wait for a response before returning `nil`. Waits forever if omitted.
+    # @return [Event, nil] The event object that was triggered, or `nil` if a `timeout` was set and no event was raised in time.
+    # @raise [ArgumentError] if `timeout` is given and is not a positive numeric value
+    def add_await!(key, type, attributes = {})
+      raise "You can't await an AwaitEvent!" if type == Discordrb::Events::AwaitEvent
+
+      timeout = attributes[:timeout]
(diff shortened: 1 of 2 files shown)


z64:
I've pushed some updates to this after some thought and talking a bit.

- **This no longer wraps the exisiting await system.** It now registers a handler itself an un-registers afterwards via `register_event` and `remove_handler`.
- The next most common use case in awaits is having them "expire" after a while. I really like this, and have added a "reserved" `timeout:` attribute that will signal the CV if an event is not received in time.

Now we have this:
```rb
require 'discordrb'

bot = Discordrb::Bot.new token: ENV['TOKEN']

bot.message(content: '!test') do |event|
  event.respond 'What is your name?'
  response = event.message.await!(timeout: 3)
  if response
    event.respond "Hello, #{response.message.content}!"
  else
    event.respond 'You took too long!'
  end
end

bot.run
```

z64:
Ahhh, caught by the new YARD CI :laughing:

PR #503, opened 2018-05-27
Update required Ruby version in the README.

connorshea:
The required Ruby version in the gemspec is 2.2.4 or above, this was changed with 4030703b18e15394936d48568a10de182fab665b.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ See also: [Documentation](http://www.rubydoc.info/gems/discordrb), [Tutorials](h
 
 ## Dependencies
 
-* Ruby 2.2.4+
+* Ruby 2.2+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ See also: [Documentation](http://www.rubydoc.info/gems/discordrb), [Tutorials](h
 
 ## Dependencies
 
-* Ruby 2.1+
+* Ruby 2.2.4+
```
z64: Should we really be specifying patch version? Ruby apparently has been following semver since 2.1

https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/

Our CI will also be using the latest patch

PR #505, opened 2018-06-03
Add attr_reader :server to Role

lisinge:
Needed to know which server the role belongs to an event and found that the server isn't exposed on `Role`


z64:
Thanks!

PR #506, opened 2018-06-03
Remove name delegate to role from ServerRoleDeleteEvent

lisinge:
Having the delegate prevented the event from firing successfully.  
I don't know if there were a plan to add a temporary role to delegate too, considering this comment here https://github.com/lisinge/discordrb/blob/7a6600086c246f3bbf4283fdd149278f0b09c085/lib/discordrb/events/roles.rb#L60-L62


lisinge:
Please feel free to nitpick as i'm not at all comfortable in the codebase and with the discord API :D

z64:
Yep, this is an oversight. As that comment describes the name data is gone by the time the event is raised so it's impossible to match on it.

lib/discordrb/container.rb
```diff
@@ -397,7 +397,7 @@ def server_role_create(attributes = {}, &block)
 
     # This **event** is raised when a role is deleted.
     # @param attributes [Hash] The event's attributes.
-    # @option attributes [String] :name Matches the role name.
+    # @option attributes [String] :id Matches the role id.
```
z64: The doctype `String` needs to be updated here. You can grep the source for other places where we ask for IDs - typically we denote this as, for example, `[Role, #resolve_id]` (it may be inconsistent in places though :smile: )

Primarily the `#resolve_id` helper exists so that we can accept IDs as `String` or `Integer`, or any object that has an ID.

lisinge:
@z64 good spot, completely missed that.

I went for `#resolve_id` to stay consistent with the rest of the docs for `container.rb` as it uses just `#resolve_id`. Hope that's fine

z64:
Yep, that's fine. Eventually, some PR should be made to make it consistent across the library.. :smile:

z64:
Thanks for the contributions! :heart: 

If you have any other questions about the library or the API, feel free to join us [on Discord](https://discord.gg/cyK3Hjm).

PR #519, opened 2018-08-06
Add hello world bot to README.md

drzel:


z64:
Hi @drzel ,

I think we'll pass on this README change, as there are too many things to consider when using discordrb's API wrapper. Also, another caveat I forgot to mention: Your bot has to have connected to the gateway at least once before it will be authorized to use `create_message`, so for many, if a beginner starts with this example it will more than likely fail right away.

As I've said though, it's definitely something you should use if your use case is right. So for now I've created this wiki page that details how `Discordrb::API` works, and hopefully that will help guide others to make the right choice going forward.

https://github.com/meew0/discordrb/wiki/API-Wrapper

Thanks! :heart:

PR #529, opened 2018-09-01
Add flag to allow bot to treat edited message as if they're new messages

HorizonShadow:
I've had a lot of users ask me to implement this on my bot, and I couldn't find a built-in way to do it. Hopefully this is an acceptable way to go about it.

Basically lets people edit the message to fix typos, and still run the command. So if you have a command `:name`, and someone typed `nam`, they could edit the message and it would run the command, instead of retyping the whole thing.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -101,17 +101,20 @@ module Discordrb
     #   to Discord's gateway. `:none` will request that no payloads are received compressed (not recommended for
     #   production bots). `:large` will request that large payloads are received compressed. `:stream` will request
     #   that all data be received in a continuous compressed stream.
+    # @params parse_edited [true, false] Whether the bot should treat edited messages as a new message when 
+    #   processing commands.
     def initialize(
         log_mode: :normal,
         token: nil, client_id: nil,
         type: nil, name: '', fancy_log: false, suppress_ready: false, parse_self: false,
         shard_id: nil, num_shards: nil, redact_token: true, ignore_bots: false,
-        compress_mode: :stream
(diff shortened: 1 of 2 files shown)


z64:
Hi @HorizonShadow ! Thank you for your interest in contributing :slightly_smiling_face: 

I don't think this is a good design for us to be mixing events across event handlers of different types. Given that this is kind of a niche feature, and we just can't have an option for everything, this is a :-1: from me.

This can already be done by capturing the event handler, locating it later or just distributing it across all event handlers (requires `instance_variable_get`) like your code in this PR is doing. 

Here is an example of the first way:

```rb
require "discordrb"

bot = Discordrb::Bot.new(token: ENV["TOKEN"])

handler = bot.message(content: "!ping") do |event|
  event.respond "pong!"
end

bot.message_edit do |event|
  if handler.matches?(event)
    handler.call(event)
  end
end

bot.run
```

HorizonShadow:
Thanks so much for the alternative implementation! I was kind of at a loss of how to do thos easily within my bot, which lead to the PR so this helps a lot.

If you think it’s a niche feature, I’m not going to argue, and I can definitely see the problem with registering message events in an edited message routine. 

Thanks again for the ideas for how to do this on my side!

z64:
No problem. If you're ever stuck and need guidance on whether you have to "get your hands dirty", you can of course just run it by us in the discord channel. (https://discord.gg/cyK3Hjm)
Sometimes you might have to modify the lib, sometimes our documentation isn't clear, or whatnot.

re: niche
I usually go by my experience in the channel; I've been using drb for about 2 years and never heard someone ask for this :)

PR #530, opened 2018-09-01
Add "Priority Speaker" as bit 8

Chewsterchew:
Recently Discord added a permission value for 256, that being Priority speaker.
Check this example
image
Not sure how useful this could be or if I'm missing anything, but I just saw it wasn't even there.


z64:
Thanks!

PR #577, opened 2018-10-16
Improve wording of server streaming timeout message

connorshea:
This can timeout because the JSON payload for a given server is taking a long time to send, so the previous warning was confusing.


connorshea:
Closing this since #578 supersedes it, and was approved.

PR #579, opened 2018-10-26
Implement Paginator utility

z64:
This PR adds `Paginator`, a utility class for wrapping paginated requests, and implements it right away in `Channel#history`. 

This primarily addresses the FAQ of users who want to retrieve more than 100 messages in a channel, and find it difficult to get the loop logic correct.

Now you can do this:

```rb
array = channel.history(1000).to_a
array.size #=> 1000
```

and discordrb will make as many requests as needed to fetch 1000 messages. Additionally you can pass `nil` to `amount` to fetch all of a channel's history in either direction.

`Paginator` is `Enumerable` so most methods you would already be using on `Array<Message>` will still work:

```rb
# Fetch the last 1000 messages, and select messages from z64
array = channel.history(1000).select do |message|
  message.author.name == "z64"
end
```

There are other requests we can use `Paginator` for, but those parts of the library aren't well prepared for that. They can be added in a later PR; `Channel#history` is the most used paginated endpoint.

## Compatibility

- `Channel#history` now takes named arguments. This is purely a personal vendetta, as I can never remember the order of the arguments. I can be persuaded to have both (named & positional) for a release, but I doubt I am the only one that has struggled with this, so I think this change is fine.

- If requesting 100 or fewer messages, or when using `around_id`, the behavior is the same; `Array<Message>` is returned. People who request more than 100 messages will get the new `Paginator`. Later I would like to always return a paginator, regardless.

- When using a `Paginator` request, messages are sorted *in the direction* of the request. Using `after_id` will return messages oldest to newest. Using `before_id` will return messages newest to oldest. I think it would be too complicated and slow to always return in the same order, as for one direction you would have to buffer every page before you could start giving messages to the user, who might not even need to request every page (i.e., they `break`). They primary use case for something like scraping large sections of history anyways should be for storing or updating a database of messages from downtime, in which the yielded order does not matter.

## (ex-) Guild Member Rate limiting

A last note is that when you roll over channel history that has messages from members that left the server, we will spam the API with requests for that member for each message encountered; this will indeed slow things down. We can avoid this, but the caching logic in `Message#initialize` is too complicated to change in one PR. I'm fine with leaving it as an optimization to make at a later date. Of course, this was already the case before this PR - but this PR makes that issue more efficiently exploited, so users may report it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `Server#bot` method for obtaining your bot's own `Member` on a particular server ([#597](https://github.com/meew0/discordrb/pull/597))
 - `Attachment#spoiler?`, to check if an attachment is a spoiler or not ([#603](https://github.com/meew0/discordrb/pull/603), thanks @swarley)
 - Methods on `Server` to manage the server's emoji ([#595](https://github.com/meew0/discordrb/pull/595), thanks @swarley)
+- `Paginator` utility class for wrapping paginated endpoints ([#579](https://github.com/meew0/discordrb/pull/579))
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


z64:
After some discussion, we're going to merge the Paginator utility **without** the changes to `history`, which may come in a later PR.

PR #582, opened 2018-10-29
Fix Ruby version requirement in README

ChallahuAkbar:
Just realized I forgot this in the previous PR :smiley: (#487)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ See also: [Documentation](https://www.rubydoc.info/gems/discordrb), [Tutorials](
 
 ## Dependencies
 
-* Ruby >= 2.3.7
+* Ruby >= 2.3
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,7 +23,7 @@ See also: [Documentation](https://www.rubydoc.info/gems/discordrb), [Tutorials](
 
 ## Dependencies
 
-* Ruby 2.2+
+* Ruby >= 2.3.7
```
z64: I think in general, we should be dropping specifications for patch versions of dependencies everywhere unless they are important (i.e., CVE) or we otherwise rely on a fix in that release

```suggestion
* Ruby >= 2.3
```

z64: @ChallahuAkbar 🏓 ?

PR #589, opened 2018-12-27
Add MessageEvent#await!

soukouki:
```ruby
bot.command :test do |event|
  # before
  res = event.message.await!
  # after
  res = event.await!
end
```

I didn't add `MessageEvent#await`. The reason is `await` will be deprecated.

### DIFF
--- a/lib/discordrb/events/message.rb
+++ b/lib/discordrb/events/message.rb
@@ -139,6 +139,12 @@ module Discordrb::Events
       @file = nil
     end
 
+    # Add a blocking {Await} for a message with the same message author and channel.
+    # @see Bot#add_await!,
+    def await!(attributes = {})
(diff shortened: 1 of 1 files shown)


VxJasonxV:
I can't think of a situation why you would await on an event, semantically. Once the event triggers, there's nothing to re-trigger it. `await`ing a message has some context like reactions, edits, deletes, etc.

The only context I see is the doc addition line:
>      # Add a blocking {Await} for a message with the same message author and channel.

But any given event is way more than the same message author and channel, because the raw event can be a server change, member state change (join, leave, status change), voice channel state change, reaction, etc.

Based off the change you described in this Pull Request's description it seems like simplification for the sake of simplification, but without the correct semantic purpose. (That's a bad thing.)

soukouki:
I agree.

Thank you for detailed explanation.

PR #593, opened 2019-01-15
Validate bot token using API.gateway_bot before connecting

z64:
This uses the authenticated `API.gateway_bot` endpoint to retrieve the gateway URL in `Gateway#find_gateway`. If this step fails with an uncaught exception, and we cannot retrieve a gateway URL, this is because the bot's token is incorrect. The token can not be corrected at runtime, so we abort the connection process.

`Bot#type` is exposed to be used internally to decide if we should validate the token in this manner.

### DIFF
--- a/lib/discordrb/api.rb
+++ b/lib/discordrb/api.rb
@@ -290,6 +290,18 @@ module Discordrb::API
     )
   end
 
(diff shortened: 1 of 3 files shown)


z64:
This is too heavy handed.

PR #594, opened 2019-01-16
Add options to disable case-sensitivity for commands and not have bot retry on 502 errors

dheimerl:


z64:
I've seen this behavior before, extremely rarely. The thing is, the 502 tells us our request never reached Discord. This is served by Cloudflare. It's important for the library to be as resilient as we can to network failures when they don't reach Discord. Additionally, other libraries (with *much* larger user bases) follow this same behavior - unconditional retry on 502; so I don't think we will change this for everyone, nor can we just add an option for everything, especially for such a rare problem.

You can still change this if you like by simply patching the `retry` out of `raw_request`. Bear in mind this API is reworked in #539, which will be merged eventually once other blocking issues move forward.

PR #598, opened 2019-01-21
Test Circle configuration

z64:


z64:
Oops. Didn't mean to open this here (yet). Still exploring this option.

PR #605, opened 2019-02-21
send_file filename

swarley:
# Summary

Add the ability to provide a filename that will override the original name of the file when it is sent.

```ruby
event.send_file(File.open('rubytaco.png', 'r'), filename: 'the_ruby_taco.png')
```

---

## Added

Specs for `Bot#send_file`
Specs for `MessageEvent#attach_file`

## Changed

Added `filename` kwarg to the following:

- `Bot#send_file`
- `Channel#send_file`
- `User#send_file`
- `MessageEvent#send_file`
- `MessageEvent#attach_file`

### DIFF
--- a/lib/discordrb/events/message.rb
+++ b/lib/discordrb/events/message.rb
@@ -82,8 +82,11 @@ module Discordrb::Events
     # @return [String] the message that has been saved by calls to {#<<} and will be sent to Discord upon completion.
     attr_reader :saved_message
 
-    # @return [File] the file that have been saved by calls to {#attach_file} and will be sent to Discord upon completion.
+    # @return [File] the file that has been saved by a call to {#attach_file} and will be sent to Discord upon completion.
     attr_reader :file
+
+    # @return [String] the filename set in {#attach_file} that will override the original filename when sent.
+    attr_reader :filename
 
     # @!attribute [r] author
     #   @return [Member, User] who sent this message.
@@ -110,6 +113,7 @@ module Discordrb::Events
       @channel = message.channel
       @saved_message = ''
       @file = nil
+      @filename = nil
     end
 
     # Sends file with a caption to the channel this message was sent in, right now.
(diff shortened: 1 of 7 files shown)


spec/bot_spec.rb
```diff
@@ -136,4 +136,30 @@
...
+    end
+
+    it 'does not define original_filename when filename is nil' do
+      file = double(:file)
+
+      allow(file).to receive(:respond_to?).with(:read).and_return(true)
+      allow(Discordrb::API::Channel).to receive(:upload_file).and_return('{}')
+      allow(Discordrb::Message).to receive(:new)
+
+      expect(file).not_to receive(:define_singleton_method)
```
z64: Actually, I'd rather we work over doubles and not expect metaprogramming:
```rb
original_filename = double(:filename)
file = double(:file, original_filename: original_filename, read: true)
new_filename = double("new filename")

```
Then in either spec:
```rb
expect(file.original_filename).to eq new_filename
expect(file.original_filename).to eq original_filename
```

spec/event_spec.rb
```diff
@@ -100,6 +100,39 @@
...
 
+  describe Discordrb::Events::MessageEvent do
+    let(:bot) { double }
+    let(:message) { double('message', channel: nil) }
+
+    subject :event do
+      described_class.new(message, bot)
+    end
+
+    describe '#attach_file' do
```
z64: Per my other comments, `Respondable` can be reduced to delegations to `Bot#send_file`, and these tests won't be necessary.

lib/discordrb/events/message.rb
```diff
@@ -117,20 +117,23 @@ def initialize(message, bot)
...
     # Attaches a file to the message event and converts the message into
     # a caption.
     # @param file [File] The file to be attached
-    def attach_file(file)
+    # @param filename [String] Overrides the filename of the uploaded file
+    def attach_file(file, filename: nil)
       raise ArgumentError, 'Argument is not a file!' unless file.is_a?(File)
 
       @file = file
+      @file.define_singleton_method(:original_filename) { filename } if file.respond_to?(:read) && filename
```
z64: `@file` eventually makes its way to `Bot#send_file` in the `after_call` method. Instead of duplicating behavior, how about forwarding `@filename`?

spec/event_spec.rb
z64: If you make my other tweak (`@filename`) the test to add here if you wanted would be that after calling `#attach_file`, expect that `#after_call` calls `#send_file` with our passed file & filename

PR #612, opened 2019-03-18
Adding MessageUpdateEvent

### LINKED ISSUE #507: Raise events on all MESSAGE_UPDATEs
Discord's `MESSAGE_UPDATE` events are only guaranteed to contain two keys, so a valid `MESSAGE_UPDATE` could be:
```json
{
  "id": "123",
  "channel_id": "456"
}
``` 
Currently, there are two scenarios when a `MESSAGE_UPDATE` is fired:

- A. When a message is edited (`content`, or `embeds` is changed). Most of the time this contains many additional fields.
- B. When Discord itself modifies the message (appending embeds to the message from URLs detected in `content`)

In the case of A, many additional fields are present, and we detect the presence of the `author` field and raise a `MessageEditEvent`. We otherwise *ignore* the event if `author` is not present.

In the case of B, *only* the updated fields will be present. When embeds are attached, the `author` field won't be there, so its currently impossible to handle this kind of event without patching the library's internal event handler method to raise `MessageEditEvent` even if the author is missing:

```rb
class Discordrb::Bot
  def update_message(data)
    message = Discordrb::Message.new(data, self)
    return if message.author # (Prevents the event from being raised twice)
    event = Discordrb::Events::MessageEditEvent.new(message, self)
    raise_event(event)
  end
end
```

We should support both of these use cases by providing `MessageUpdateEvent` via `bot.message_update`, which is *always* raised regardless of the updated message's structure, and let the user handle it however they want. This way it is not a breaking change, and users can rely on `event.author` when listening to `bot.message_edit`.

The dispatch case should be handled to something similar to:
```rb
when :MESSAGE_UPDATE
  update_message(data)

  message = Message.new(data, self)
  if message.edited_timestamp
    event = MessageEditEvent.new(message,self)
    raise_event(event)
  end

  event =  MessageUpdateEvent.new(message,self)
  raise_event(event)
```

Perhaps `edited_timestamp` is a more reliable field to act upon, or maybe `content` would make more sense. Further testing should be done to see for sure.

---

[Example event sequence of posting an imgur link and discord embedding it](https://gist.githubusercontent.com/z64/40dcb698a7ce5e84215535d201ca39a7/raw/89d45b8d75a8b1adba684c5c3a090753e8ec3286/message_update.txt)

Ref #95 , that added this filtering on `author`

swarley:
# Summary

Add a handler that fires on every `MESSAGE_UPDATE`.
Closes #507

## Added

`Events::MessageUpdateEvent`
`Events::MessageUpdateEventHandler`
`EventContainer#message_update`

## Changed

`Bot#handle_dispatch`

### DIFF
--- a/lib/discordrb/container.rb
+++ b/lib/discordrb/container.rb
@@ -109,6 +109,18 @@ module Discordrb
       register_event(MessageDeleteEvent, attributes, block)
     end
 
+    # This **event** is raised whenever a MESSAGE_UPDATE. This event catches some message updates that may
+    # not raise a MessageEditEvent. See [#507](https://github.com/meew0/discordrb/issues/507) for details.
+    # @param attributes [Hash] The event's attributes.
+    # @option attributes [#resolve_id] :id Matches the ID of the message that was updated.
+    # @option attributes [String, Integer, Channel] :in Matches the channel the message was updated in.
+    # @yield The block is executed when the event is raised.
+    # @yieldparam event [MessageUpdateEvent] The event that was raised.
+    # @return [MessageUpdateEventHandler] the event handler that was registered.
+    def message_update(attributes = {}, &block)
+      register_event(MessageUpdateEvent, attributes, block)
+    end
+
     # This **event** is raised when somebody reacts to a message.
     # @param attributes [Hash] The event's attributes.
     # @option attributes [Integer, String] :emoji Matches the ID of the emoji that was reacted with, or its name.
--- a/lib/discordrb/container.rb
+++ b/lib/discordrb/container.rb
@@ -109,8 +109,10 @@ module Discordrb
       register_event(MessageDeleteEvent, attributes, block)
     end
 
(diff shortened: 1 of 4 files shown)


lib/discordrb/container.rb
```diff
@@ -109,6 +109,18 @@ def message_delete(attributes = {}, &block)
       register_event(MessageDeleteEvent, attributes, block)
     end
 
+    # This **event** is raised whenever a MESSAGE_UPDATE. This event catches some message updates that may
+    # not raise a MessageEditEvent. See [#507](https://github.com/meew0/discordrb/issues/507) for details.
```
z64: This handler can be explained without needing to refer to a github issue. Try something like:

> This **event** is raised whenever a message is updated. Message updates can be triggered from a user editing
> their own message, or from Discord automatically attaching embeds to the user's message for URLs
> contained in the message's content. If you only want to listen for users editing their own 
> messages, use the `message_edit` handler instead.

PR #616, opened 2019-03-24
Remove #resolve_id duck typing

### LINKED ISSUE #512: Consistent typedocs for #resolve_id arguments
Across our API, we have a lot of methods that accept an object that responds to `#resolve_id`.

`resolve_id` is a helper method defined on `String`, `Integer`, and `IDObject`.

```rb
"123".resolve_id # => 123
123.resolve_id # => 123

class User
  include IDObject
  # ..
end

# Get a User object somehow
user.resolve_id # => 123
```

This allows users freedom to pass instances of our data objects (any object that has a Discord ID) as arguments, or the IDs as `String` or `Integer`.

```rb
server.ban(97437477723193344)
server.ban('97437477723193344')
user = bot.user(97437477723193344)
server.ban(user)
```

Our resulting typedocs are inconsistent, sometimes they will list the type, as well as Integer and String, followed by `#resolve_id`.

[See sample grep](https://gist.github.com/z64/7d95e44df0addfd44f33e44645508910)

My proposal is that they should all be unified like so:

```rb
# @param foo [Type, #resolve_id] the Type of ID of a type to do bar
```

and users should see the documentation for `#resolve_id` to understand what can be passed.

There may be exceptions, but I'll leave that to anyone willing to dig into this and investigate to bring up and we can discuss further.

swarley:
# Summary

Remove all instances of `#resolve_id` duck typing in docs. Clarify that an ID can be used in many places instead of the relevant object.

Many docs types are changed as well as inclusion of the mention of ID in the description string. Many of these could be potentially worded differently.

Closes #512

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -339,7 +339,7 @@ module Discordrb
 
     # Disconnects the client from a specific voice connection given the server ID. Usually it's more convenient to use
     # {Discordrb::Voice::VoiceBot#destroy} rather than this.
-    # @param server [Server, String, Integer] The server, or server id, the voice connection is on.
+    # @param server [Server, String, Integer] The server, or server ID, the voice connection is on.
     # @param destroy_vws [true, false] Whether or not the VWS should also be destroyed. If you're calling this method
     #   directly, you should leave it as true.
     def voice_destroy(server, destroy_vws = true)
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -162,7 +162,7 @@ module Discordrb
 
     # @overload emoji(id)
     #   Return an emoji by its ID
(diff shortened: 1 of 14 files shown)


lib/discordrb/bot.rb
```diff
@@ -339,7 +339,7 @@ def voice_connect(chan, encrypted = true)
 
     # Disconnects the client from a specific voice connection given the server ID. Usually it's more convenient to use
     # {Discordrb::Voice::VoiceBot#destroy} rather than this.
-    # @param server [Server, Integer, #resolve_id] The server the voice connection is on.
+    # @param server [Server, String, Integer] The server, or server id, the voice connection is on.
```
connorshea: Capitalize it consistently reeeee
```suggestion
    # @param server [Server, String, Integer] The server, or server ID, the voice connection is on.
```

PR #618, opened 2019-03-30
Add news? and store?

swarley:
# Summary

Add checks for the new channel types, `store` and `news`.

---

## Added

`Discordrb::Channel#news?`
`Discordrb::Channel#store?`

`{news: 5, store: 6}` to `TYPES` constant


z64:
Granted they are only available for storefront guilds, maybe we should update the channel type whitelist in `Server#create_channel`? I think I'd rather let Discord return an error than make a qualified user have to use `API::Server` to create one.

PR #620, opened 2019-04-02
Adding super fast parse_mentions

swarley:
# Summary

Honestly, ever since I started programming in ruby, I missed the wind blowing through my hair that only C can provide. This is a much needed upgrade and the proper step towards the blazing fast discordrb that I know we can achieve.

## Added

`parse_mentions.c`

## Changed

Slow mode is now fast mode

## Deprecated

Ruby slowness related injuries.

## Removed

The feeling of slow ruby

## Fixed

`Bot#parse_mentions` - now achieves the speed that i crave

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_script:
   - chmod +x ./cc-test-reporter
   - ./cc-test-reporter before-build
 script:
+  - bundle exec rake compile
   - bundle exec rspec spec
   - bundle exec rubocop -c .rubocop.yml
   - bin/travis_build_docs.sh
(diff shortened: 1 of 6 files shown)


SanksTheYokai:
> I have no power here but I’ll approve it anyway
I have no power here too, but shouldn't we save slower option to let it run on another ruby interpretations that do not support C bindings? Like jruby

SanksTheYokai:
Slower option too, sorry

z64:
https://www.youtube.com/watch?v=1Npo0cmp-VY

PR #625, opened 2019-04-07
Fix VoiceStateUpdateEvent handler matching

swarley:
# Summary

~~Remove unused accessor, clean up code, and~~ fix logic in handler matching.

## ~~Removed~~

~~Removed the accessor referencing `@token` in `VoiceStateUpdateEvent`.~~

## Fixed

`VoiceStateUpdateEventHandler#matches?` - Change matching that was previously using inappropriate logic.


swarley:
Isolated to only fix faulty matching logic. Style will be ignored and removing the invalid instance variable can be addressed at a later time

PR #630, opened 2019-04-23
fix message#timestamp returning non-local timezone

jacob-thompson:
# Summary

Message#timestamp was returning a time that was in a specific timezone.
More specifically, `data['timestamp']` returns a non-local time, therefore my solution was to work around using this data.
I fixed this by using `Time.now` rather than `Time.parse(data['timestamp'])` in the Message class.
There is likely a better way to go about this fix, though my only goal was to fix this specific issue.

## Screenshots

[Before this change](https://imgur.com/ZBBUXSj)
[After this change](https://imgur.com/J0s1u0w)

## Changes

in discordrb/lib/discordrb/data/message.rb:
from
`@timestamp = Time.parse(data['timestamp']) if data['timestamp']`
to
`@timestamp = Time.now if data['timestamp']`

### DIFF
--- a/lib/discordrb/data/message.rb
+++ b/lib/discordrb/data/message.rb
@@ -106,7 +106,7 @@ module Discordrb
 
       @webhook_id = data['webhook_id'].to_i if data['webhook_id']
 
-      @timestamp = Time.parse(data['timestamp']) if data['timestamp']
+      @timestamp = Time.now if data['timestamp']
       @edited_timestamp = data['edited_timestamp'].nil? ? nil : Time.parse(data['edited_timestamp'])
       @edited = !@edited_timestamp.nil?
       @id = data['id'].to_i


VxJasonxV:
`Time.now` uses the time, well, now. This method is reading in data about a message, you want the time of the message, not the time that it was processed.

1. If you're worried about the timestamp, convert it.
2. The 22 second delay is probably happening because your machine's clock (or wherever the bot is running) is 22 seconds out of sync. Sync your clock.

jacob-thompson:
> `Time.now` uses the time, well, now. This method is reading in data about a message, you want the time of the message, not the time that your bot processed it.
> 
> 1. If you're worried about the timezone, convert it.
> 2. The 22 second delay is probably happening because your machine's clock (or wherever the bot is running) is 22 seconds out of sync. Sync your clock.

why the fuck would u want the timestamp of a message in ur server to be in a different timezone to begin with? what is the problem with making the timezone local in the first place? what is the point of going thru the extra effort of converting the timezone each time you call `message#timestamp`? genuinely curious. also im pretty sure `data['timestamp']` returning an inaccurate time has nothing to do with my machine lOL

z64:
Hi @jacob-thompson,

First, thank you for your interest in contributing back to discordrb! While its great you want to contribute, I *will not tolerate* that language and attitude on our tracker. Please refrain from doing so in the future - Jason brings up a valid point, but besides that, he's just trying to help, like you. Thank you!

---

Back on topic of this PR -

This change isn't quite right. We cannot simply throw away the time information that Discord gave to us - when your bot received the message is *not* the same time as when the message was sent. It could, in fact, be any amount of time. So, it is entirely expected for it to not match your machines time, and is actually Discord's server time, which, you should be synced with via `ntp`, etc  -  **if you are not**, this will *indeed* skew a subtraction if your clock is not synced, similar to what is shown in your screenshot. A lot of users run into this issue and get a "negative ping", exactly like you did.

Secondly, this would be broken for other kinds of event that wrap a Message object. For example, if a message is edited, `event.message.timestamp` would be the current time, instead of when the message was originally created! This would certainly be an unexpected bug.

More broadly, the fact that `event.timestamp` is a non-local timezone is not an issue, as `Time` instances are correctly comparable across timezones.

z64:
Per the above, this PR as it is introduces bugs and it isn't clear what issue it solves. If anything I mentioned wasn't clear, feel free to comment back and/or join us in Discord where we can work more closely with you to identify whether there was something actionable here.

Thank you for your interest in contributing!

PR #631, opened 2019-04-26
Add dispatch timeout attribute to the bot

einzige:
# Summary

There was hardcoded 10 seconds condition which caused bot to reconnect
too many times (causes bot to be banned on Discord). The solution is to
give it more time if it's really needed.

It started from here:
image

## Added

`attr_reader :dispatch_timeout` to `Discordrb::Bot` so we can use:

```ruby
Discordrb::Bot.new(dispatch_timeout: 100) # replaced default 10 seconds
```

## Changed

Nothing

## Deprecated

Nothing

## Removed

Nothing

## Fixed

Nothing

-------

Rubocop, RSpec passed.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -62,6 +62,10 @@ module Discordrb
     # @return [Gateway] the underlying {Gateway} object.
     attr_reader :gateway
 
+    # The number of seconds to wait for unavailable servers since READY
+    # @return [Integer]
+    attr_reader :dispatch_timeout
+
     include EventContainer
     include Cache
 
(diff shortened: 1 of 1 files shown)


z64:
Sorry, I don't understand this change. That timeout mechanism does not cause any re-connection. It just calls your `bot.ready` handler early if it looks like we might be waiting a long time, and conditionally emits a warning if you try to access cached data before its fully available.

swarley:
If this fixed a problem you were having could you please open an issue describing the nature of it?

I'm interested in what the context of this is.

z64:
It's still unclear what issue this PR addresses per above. If you have further information or questions, I highly recommend asking in Discord where we can work more closely with you to debug any issues you might be having.

Thank you for your interest in contributing!

PR #637, opened 2019-06-14
Add #create_webhook method to create a webhook in a channel

Chew:
# Summary

When I was making webhooks for a channel, I found there's no easy way to do this. So, using the power of my somewhat limited Ruby knowledge (help me out here), I was able to figure out how to do it! So, I wrote this method. It returns a Webhook object.

image

The method takes in 3 parameters. 1 required (*name* of the webhook), 2 optional (*avatar* of the webhook, and *reason* for adding it for the Audit Log)

---

## Added

* #create_webhook method to the Channel object.

### DIFF
--- a/lib/discordrb/data/channel.rb
+++ b/lib/discordrb/data/channel.rb
@@ -704,7 +704,8 @@ module Discordrb
     # @param reason [String] the reason why this happened, for the audit logs
     # @return [Webhook] the instantiated webhook.
     def create_webhook(name, avatar = nil, reason = nil)
-      raise 'Tried to create a webhook in a non-server channel' unless server
+      raise ArgumentError, 'Tried to create a webhook in a non-server channel' unless server
+      raise ArgumentError, 'Tried to create a webhook in a non-text channel' unless text_channel?
 
       hook = API::Channel.create_webhook(@bot.token, @id, name, avatar, reason)
       Webhook.new(JSON.parse(hook), @bot)
--- a/lib/discordrb/data/channel.rb
+++ b/lib/discordrb/data/channel.rb
@@ -700,15 +700,15 @@ module Discordrb
 
     # Creates a webhook in this channel
     # @param name [String] the default name of this webhook.
-    # @param avatar [String] the default avatar URL to give this hook
-    # @param reason [String] the reason why this happened, for the audit logs
-    # @return [Webhook] the instantiated webhook.
+    # @param avatar [String] the default avatar URL to give this webhook.
+    # @param reason [String] the reason for the webhook creation.
+    # @return [Webhook] the created webhook.
     def create_webhook(name, avatar = nil, reason = nil)
       raise ArgumentError, 'Tried to create a webhook in a non-server channel' unless server
       raise ArgumentError, 'Tried to create a webhook in a non-text channel' unless text_channel?
 
-      hook = API::Channel.create_webhook(@bot.token, @id, name, avatar, reason)
-      Webhook.new(JSON.parse(hook), @bot)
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/channel.rb
```diff
@@ -698,6 +698,18 @@ def leave_group
 
     alias_method :leave, :leave_group
 
+    # Creates a webhook in this channel
+    # @param name [String] the default name of this webhook.
+    # @param avatar [String] the default avatar URL to give this hook
+    # @param reason [String] the reason why this happened, for the audit logs
+    # @return [Webhook] the instantiated webhook.
+    def create_webhook(name, avatar = nil, reason = nil)
+      raise 'Tried to create a webhook in a non-server channel' unless server
```
z64: Please raise `ArgumentError` (see for example `Channel#category=`).

Could we also check `unless server && text_channel?` and update the message? They could be separate error checks, or just one - I don't feel strongly either way.

lib/discordrb/data/channel.rb
```diff
@@ -698,6 +698,18 @@ def leave_group
...
+
+      hook = API::Channel.create_webhook(@bot.token, @id, name, avatar, reason)
+      Webhook.new(JSON.parse(hook), @bot)
```
Chew: Is this a good place for the JSON.parse, btw? Or should it be on the line where hook is defined?

z64: We most commonly do:

```rb
response = API.foo(...)
Obj.new(JSON.parse(response))
```

lib/discordrb/data/channel.rb
Chew: Added this, just did 2 rescues. 
```ruby
raise ArgumentError, 'Tried to create a webhook in a non-server channel' unless server
raise ArgumentError, 'Tried to create a webhook in a non-text channel' unless text_channel?```

However, the #webhooks method does not raise ArgumentError, 
```ruby
raise 'Tried to request webhooks from a non-server channel' unless server```

Not sure if it should or not, but I replicated my original rescue from the #webhook method.

Comment (z64) [reply lib/discordrb/data/channel.rb]: That's fine. We'll consider it out of scope for this PR.

File: lib/discordrb/data/channel.rb [hunk 3]
```diff
@@ -698,6 +698,19 @@ def leave_group
 
     alias_method :leave, :leave_group
 
+    # Creates a webhook in this channel
+    # @param name [String] the default name of this webhook.
+    # @param avatar [String] the default avatar URL to give this hook
```
z64:
```suggestion
    # @param avatar [String] the default avatar URL to give this webhook
```

lib/discordrb/data/channel.rb
```diff
@@ -698,6 +698,19 @@ def leave_group
...
+    # @param reason [String] the reason why this happened, for the audit logs
```
z64: Consistent language with other `@param reason` docstrings:
```suggestion
    # @param reason [String] the reason for the webhook creation.
```

lib/discordrb/data/channel.rb
```diff
@@ -698,6 +698,19 @@ def leave_group
...
+    # @return [Webhook] the instantiated webhook.
```
z64: Consistent language with other `@return` docstrings for methods that create things:
```suggestion
    # @return [Webhook] the created webhook.
```

lib/discordrb/data/channel.rb
z64: Sorry if I wasn't clear - do you mind applying this change? (s/hook/response)

PR #640, opened 2019-07-10
Support disabling guild subscription events

z64:
Todo:

- [ ] Verification / testing
- [ ] Sufficient warning in docs


z64:
This is extremely broken.

z64:
The dangerous behavior interacting with disabling guild subscriptions was likely fixed with this commit:

https://github.com/discordrb/discordrb/commit/830dcf10d58c443259935df0ddfca389fed6033b

Either way though, traffic control will be done with intents, which supersedes the guild subscriptions toggle in the API.
No sense for us to support both, in this case.

PR #642, opened 2019-07-13
Fix to_reaction not behaving well for unicode emoji

z64:


Daniel-Worrall:
I think this needs updating to work with animated emoji

z64:
@Daniel-Worrall No; the reaction API does not care if an emoji is animated or not. The reaction format is the same either way.

Daniel-Worrall:
Yes, boss! 👍

PR #643, opened 2019-07-18
Return `nil` on 204 response

swarley:
# Summary

Currently, `Server#add_member_using_token` will raise a JSON parsing error if you attempt to add a user that is already on the server.

This is because in this case Discord returns a 204 response code with an empty body that we do not check for.

This change adds a check for an empty response body and returns `nil` if that's the case.

----------------------------------

## Changed
`Server#add_member_using_token` - Now returns `nil` when the user is already a member of the server.

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -271,12 +271,14 @@ module Discordrb
     # @param roles [Role, Array<Role, String, Integer>] the role (or roles) to give this member upon joining
     # @param deaf [true, false] whether this member will be server deafened upon joining
     # @param mute [true, false] whether this member will be server muted upon joining
-    # @return [Member] the created member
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/server.rb
```diff
@@ -271,12 +271,14 @@ def online_members(include_idle: false, include_bots: true)
     # @param roles [Role, Array<Role, String, Integer>] the role (or roles) to give this member upon joining
     # @param deaf [true, false] whether this member will be server deafened upon joining
     # @param mute [true, false] whether this member will be server muted upon joining
-    # @return [Member] the created member
+    # @return [Member, `nil`] the created member, or `nil` if the user is already a member of this server.
```
z64:
```suggestion
    # @return [Member, nil] the created member, or `nil` if the user is already a member of this server.
```

tbhockey:
🙌

Issue #581, opened 2018-10-28
Exception: #<NoMethodError: undefined method `getc' for nil:NilClass> and other fun gateway issues


oct2pus:
for what its worth i've already moved on to a Golang rewrite of my bot which does not encounter this issue, but running the original ruby version of my [bot](https://github.com/oct2pus/jadebot/tree/Original-Ruby) on a bot account with around 500 guilds or more should reproduce this issue (which is a bit of an unreasonable request i admit). Might be useful information for anyone else who runs into this issue.

---

my connection continues to reconnect in an infinite loop, I seem to get these two error messages over and over. 

```
[ERROR : heartbeat @ 2018-10-28 15:27:22.650] The websocket connection has closed: nil
[ERROR : websocket @ 2018-10-28 15:27:22.650] An error occurred in the main websocket loop!
[ERROR : websocket @ 2018-10-28 15:27:22.651] Exception: #<NoMethodError: undefined method `getc' for nil:NilClass>
[ERROR : websocket @ 2018-10-28 15:27:22.657] /home/octo/legacy/jadebot/.vendor/bundle/ruby/2.3.0/bundler/gems/discordrb-a416edf5c8be/lib/discordrb/gateway.rb:593:in `block in websocket_loop'
[ERROR : websocket @ 2018-10-28 15:27:22.658] /home/octo/legacy/jadebot/.vendor/bundle/ruby/2.3.0/bundler/gems/discordrb-a416edf5c8be/lib/discordrb/gateway.rb:593:in `synchronize'
[ERROR : websocket @ 2018-10-28 15:27:22.658] /home/octo/legacy/jadebot/.vendor/bundle/ruby/2.3.0/bundler/gems/discordrb-a416edf5c8be/lib/discordrb/gateway.rb:593:in `websocket_loop'
[ERROR : websocket @ 2018-10-28 15:27:22.658] /home/octo/legacy/jadebot/.vendor/bundle/ruby/2.3.0/bundler/gems/discordrb-a416edf5c8be/lib/discordrb/gateway.rb:569:in `connect'
[ERROR : websocket @ 2018-10-28 15:27:22.658] /home/octo/legacy/jadebot/.vendor/bundle/ruby/2.3.0/bundler/gems/discordrb-a416edf5c8be/lib/discordrb/gateway.rb:463:in `block in connect_loop'
[...]

PR #652, opened 2019-08-01
Clean up pre-2.4 code

PanisSupraOmnia:
# Summary

There were a couple spots that were missed when Ruby 2.3 support was dropped, this PR updates those spots. Theoretically the changes should also improve performance, but I haven't benchmarked and it's probably negligible.

---

## Fixed

- Use `Regexp#match?` when the `MatchData` isn't needed.
- Don't use a mutex for websocket sync because the Ruby bug that made it an issue was fixed in 2.4.


PanisSupraOmnia:
Just realized that the websocket change might be wrong since I forgot about #581. I haven't personally looked into that issue, so I'm not sure whether removing the mutex might exacerbate it or not.

z64:
#581 isn't going anywhere anyways, since we have no information / cannot reproduce. I'm fine changing this and seeing what happens.

fizvlad:
`bot#stop` method [here](https://github.com/discordrb/discordrb/blob/1cc8a4a17cc8b4f73c4af8bce01ac772bdd3682a/lib/discordrb/bot.rb#L261) is now broken

PR #654, opened 2019-08-24
Update rest-client to latest version

ali-l:
## Changed
- Bump rest-client to 2.1.0


z64:
We can actually return this to `>= 2.0.0` now that `rest-client` has released `2.1.0`. Due to RC's bumpy history, some users have reported issues with other gems in the ecosystem that require 2.0 when we were pinned to the 2.1 release client. discordrb only requires at least 2.0, so it's convenient for us to allow it.

Thanks!

PR #656, opened 2019-09-06
Parse Integer arguments in base 10

joshleblanc:
# Summary

The integer argument type doesn't parse strings into a base 10 integer. This pull request sets the radix on the parse.

Problem case:

You have a command that will delete a record from a database, based on ID. The command is "delete some_id". A user calls "delete 0066". The integer passed to the command will be `54`, because "0066" is an octal string for 54.

---

## Changed
 * command_bot.rb - Added radix to Integer(arg) call
* prefix_proc.rb - Added radix to both Integer calls


HorizonShadow:
Done

PR #662, opened 2019-10-28
Fix message split in #split_message

captainSV:
# Summary

---

## Changed
#split_message now automatically splits messages on the last ` ` character, instead of just automatically filling out all 2000 characters, because that was causing issues with breaking links or making words commit sepuku.

### DIFF
--- a/lib/discordrb.rb
+++ b/lib/discordrb.rb
@@ -55,9 +55,15 @@ module Discordrb
     # If there is no space to logically split the message on in the message (No spaces at all), it stays the same, otherwise
     # cut string on index of the last space, and put everything after that space in the next index
     # Note: please find a way to be more inclusive of 'where to logical split message', /\b/ captures the end of string
-    0.upto(ideal_ary.length-2) { |x| ideal_ary[x].rindex(/ /).nil? ? (ideal_ary[x] = ideal_ary[x])
-    : (ideal_ary[x], ideal_ary[x+1] = ideal_ary[x][0..ideal_ary[x].rindex(/ /)],
-    (ideal_ary[x][ideal_ary[x].rindex(/ /)..-1] + ideal_ary[x+1])) }
+
+    0.upto(ideal_ary.length-2) do |x|
+      if ideal_ary[x].rindex(/ /).nil? #if there is no space to break on, just leave it be
+        (ideal_ary[x] = ideal_ary[x])  #not the best way to do this, probably, but I don't personally know a better way
+      else
+        ideal_ary[x+1] = ideal_ary[x][ideal_ary[x].rindex(/ /)..-1] + ideal_ary[x+1] #shift the split bit from the string and move it one index up
+        ideal_ary[x] = ideal_ary[x][0..ideal_ary[x].rindex(/ /)] #split the string at the last space character
+      end
+    end
 
     # Slice off the ideal part and strip newlines
     rest = msg[ideal.length..-1].strip
--- a/lib/discordrb.rb
+++ b/lib/discordrb.rb
@@ -48,21 +48,9 @@ module Discordrb
     # Find the largest element that is still below the character limit, or if none such element exists return the first
     ideal = joined.max_by { |e| e.length > CHARACTER_LIMIT ? -1 : e.length }
 
-    # If it's still larger than the character limit (none was smaller than it) split it into slices with the length
-    # being the character limit, otherwise just return an array with one element
-    ideal_ary = ideal.length > CHARACTER_LIMIT ? ideal.chars.each_slice(CHARACTER_LIMIT).map(&:join) : [ideal]
-
-    # If there is no space to logically split the message on in the message (No spaces at all), it stays the same, otherwise
-    # cut string on index of the last space, and put everything after that space in the next index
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/channel.rb
```diff
@@ -380,6 +380,7 @@ def send_multiple(content)
     # Useful for sending long messages, but be wary of rate limits!
     def split_send(content)
       send_multiple(Discordrb.split_message(content))
+      nil
```
Daniel-Worrall: Why is this here?
This is breaking behaviour as it currently returns an Array of the split content.
If you're using this in a CommandBot, remember to end your block with a nil instead of expecting this message to return nil

z64: It's not breaking because this method has no documented return value.

lib/discordrb.rb
```diff
@@ -52,6 +52,13 @@ def self.split_message(msg)
     # being the character limit, otherwise just return an array with one element
     ideal_ary = ideal.length > CHARACTER_LIMIT ? ideal.chars.each_slice(CHARACTER_LIMIT).map(&:join) : [ideal]
 
+    # If there is no space to logically split the message on in the message (No spaces at all), it stays the same, otherwise
+    # cut string on index of the last space, and put everything after that space in the next index
+    # Note: please find a way to be more inclusive of 'where to logical split message', /\b/ captures the end of string
+    0.upto(ideal_ary.length-2) { |x| ideal_ary[x].rindex(/ /).nil? ? (ideal_ary[x] = ideal_ary[x])
+    : (ideal_ary[x], ideal_ary[x+1] = ideal_ary[x][0..ideal_ary[x].rindex(/ /)],
+    (ideal_ary[x][ideal_ary[x].rindex(/ /)..-1] + ideal_ary[x+1])) }
```
z64: Sorry, this isn't acceptable. Please write this code out in longer form - this is unnecessarily difficult to read and maintain :(

captainSV: Right, so i expanded the ternary and commented about what each line does.

lib/discordrb.rb
```diff
@@ -55,9 +55,15 @@ def self.split_message(msg)
     # If there is no space to logically split the message on in the message (No spaces at all), it stays the same, otherwise
     # cut string on index of the last space, and put everything after that space in the next index
     # Note: please find a way to be more inclusive of 'where to logical split message', /\b/ captures the end of string
-    0.upto(ideal_ary.length-2) { |x| ideal_ary[x].rindex(/ /).nil? ? (ideal_ary[x] = ideal_ary[x])
-    : (ideal_ary[x], ideal_ary[x+1] = ideal_ary[x][0..ideal_ary[x].rindex(/ /)],
-    (ideal_ary[x][ideal_ary[x].rindex(/ /)..-1] + ideal_ary[x+1])) }
+
+    0.upto(ideal_ary.length-2) do |x|
+      if ideal_ary[x].rindex(/ /).nil? #if there is no space to break on, just leave it be
```
swarley: Can we use `' '`, `/\s/` or `/\b/` here?

captainSV: /\s/ is probably best in my opinion
also change it to `unless` instead of `if/else`, I was being dumb when I wrote that

lib/discordrb/data/channel.rb
omnilord: It may not be documented but it most certainly would be a breaking change for anyone already using this function and using the split message array for logging purposes.

captainSV: I don't know why this is showing up as a change, considering it's already in the repo as of #629 when it was accepted

Daniel-Worrall: Seems you're behind with commits is why.

lib/discordrb.rb
```diff
@@ -48,9 +48,9 @@ def self.split_message(msg)
     # Find the largest element that is still below the character limit, or if none such element exists return the first
     ideal = joined.max_by { |e| e.length > CHARACTER_LIMIT ? -1 : e.length }
 
-    # If it's still larger than the character limit (none was smaller than it) split it into slices with the length
-    # being the character limit, otherwise just return an array with one element
-    ideal_ary = ideal.length > CHARACTER_LIMIT ? ideal.chars.each_slice(CHARACTER_LIMIT).map(&:join) : [ideal]
+    # If it's still larger than the character limit (none was smaller than it) split it into the largest chunk without
+    # cutting words apart, breaking on the nearest space within character limit, otherwise just return an array with one element
+    ideal_ary = ideal.length > 2000 ? ideal.split(/(.{,2000}\b+?)/).reject(&:empty?) : [ideal]
```
greenbigfrog: Why is `CHARACTER_LIMIT` hardcoded to 2000 here?!
```suggestion
    ideal_ary = ideal.length > CHARACTER_LIMIT ? ideal.split(/(.{,CHARACTER_LIMIT}\b+?)/).reject(&:empty?) : [ideal]
```

captainSV: Honestly? The way i was doing my algorithmic testing i didn't feel like also having a while line for declaring character limit for, the literal one line of code in the test, so i changed it to 2000.
and, evidently, forgot to change it back before pushing to my repo. That's my bad.

Issue #619, opened 2019-03-31
undefined method delete_role/delete_member when kicking a bot


localhostdotdev:
# Summary

How to reproduce:

- Have a bot
- `bot.run(:async)`
- Add to a server via permissions url: `https://discordapp.com/oauth2/authorize/?client_id=...`
- Kick bot

```
[ERROR : websocket @ 2019-03-31 14:51:40.565] Exception: #<NoMethodError: undefined method `delete_role' for nil:NilClass>
[ERROR : websocket @ 2019-03-31 14:51:40.565] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/bot.rb:905:in `delete_guild_role'
[ERROR : websocket @ 2019-03-31 14:51:40.566] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/bot.rb:1200:in `handle_dispatch'
[ERROR : websocket @ 2019-03-31 14:51:40.566] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/bot.rb:654:in `dispatch'
[ERROR : websocket @ 2019-03-31 14:51:40.566] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:714:in `handle_dispatch'
[ERROR : websocket @ 2019-03-31 14:51:40.567] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:679:in `handle_message'
[ERROR : websocket @ 2019-03-31 14:51:40.567] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:617:in `websocket_loop'
[ERROR : websocket @ 2019-03-31 14:51:40.567] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:569:in `connect'
[ERROR : websocket @ 2019-03-31 14:51:40.567] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:463:in `block in connect_loop'
[ERROR : websocket @ 2019-03-31 14:51:40.567] /.rvm/gems/ruby-head/gems/discordrb-3.3.0/lib/discordrb/gateway.rb:462:in `loop'
[...]

PR #663, opened 2019-10-28
Fix for not rescued NoMethodError on bot kick (#619)

fizvlad:
# Summary

Fix for #619 

---

## Changed
Aborting handling of `GUILD_MEMBER_DELETE` if server not found.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -883,6 +883,7 @@ module Discordrb
     def delete_guild_member(data)
       server_id = data['guild_id'].to_i
       server = self.server(server_id)
+      return unless server # Server was deleted on GUILD_DELETE. See #619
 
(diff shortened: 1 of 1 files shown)


lib/discordrb/bot.rb
```diff
@@ -883,6 +883,7 @@ def update_guild_member(data)
     def delete_guild_member(data)
       server_id = data['guild_id'].to_i
       server = self.server(server_id)
+      return unless server # Server was deleted on GUILD_DELETE. See #619
```
z64: This code is pretty self-explanatory to most readers of the internals - the comment isn't necessary. It's well known that `bot.server(id)` can return `nil` on a cache miss, which could be for any reason, including from processing a `GUILD_DELETE`.
```suggestion
      return unless server
```

PR #667, opened 2019-11-04
Mobile Indicator

swarley:
if b1nzy doesn't correct himself in 15 minutes we're legally allowed to merge this 

image


lib/discordrb/gateway.rb
```diff
@@ -277,7 +279,7 @@ def identify
       compress = @compress_mode == :large
       send_identify(@token, {
                       '$os': RUBY_PLATFORM,
-                      '$browser': 'discordrb',
+                      '$browser': @mobile ? 'Discord Android' : 'discordrb',
```
Chew: does this have to be discord android?? seems like we're lying here rg

z64:
Discord won't be supporting this.

PR #670, opened 2019-11-09
Add User#custom_status

swarley:
## Added
`User#custom_status`
`CustomStatus` class

## Changed
`User#update_presence`

### DIFF
--- a/lib/discordrb/data/user.rb
+++ b/lib/discordrb/data/user.rb
@@ -43,6 +43,21 @@ module Discordrb
     end
   end
 
+  # Data type for custom statuses
+  class CustomStatus
+    # @return [String] the custom status message
+    attr_reader :state
+    alias_method :text, :state
+
+    # @return [Emoji, String, nil] the emoji used in the status
+    attr_reader :emoji
+
+    def initialize(data, bot)
+      @state = data['state'] || ''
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/user.rb
```diff
@@ -122,6 +140,10 @@ def update_username(username)
     def update_presence(data)
       @status = data['status'].to_sym
 
+      @custom_status = if (status_data = data['activities'].find { |act| act['type'] == 4 })
```
z64: Is this right? The `activities` field exists to "supersede" the `game` field - it is just to support *multiple* activity objects, where `game` is just one. In other words, semantically, `game` would otherwise be called "`activity`". Are you certain that that type 4 is *only* exposed in `activities`, and never in `game`, when a custom status is their *only* activity?

swarley: It's exposed in both. But it is most reliable to pull from activities in my testing. If a user has a custom status set and then enters a game, the custom status is overridden in that field. The custom status will always be present in the activity list however.

z64: Yeah I guess. This class needs to break in 4.x to get rid of the concept/methods. of a single activity.

swarley:
Closed in favor of #677

PR #671, opened 2019-11-09
Emoji fixup

swarley:
## Summary
The `server` param is treated as optional within the Emoji constructor. This PR adds a `nil` default to reflect that. Also fix some issues I introduced when adding server emoji management. That one sure flew under the radar for a while

## Change
`Bot#parse_mentions`
`Emoji#initialize`
`Server#add_emoji`
`Server#edit_emoji`
`Server#process_emoji`
`ReactionEvent#initialize`

### DIFF
--- a/lib/discordrb/data/emoji.rb
+++ b/lib/discordrb/data/emoji.rb
@@ -8,18 +8,17 @@ module Discordrb
     # @return [String] the emoji name
     attr_reader :name
 
-    # @return [Server] the server of this emoji
+    # @return [Server,] the server of this emoji
     attr_reader :server
 
-    # @return [Array<Role>] roles this emoji is active for
+    # @return [Array<Role>, nil] roles this emoji is active for
     attr_reader :roles
 
     # @return [true, false] if the emoji is animated
     attr_reader :animated
     alias_method :animated?, :animated
 
-    def initialize(data, bot, server = nil)
-      @bot = bot
+    def initialize(data, server = nil)
       @roles = nil
 
       @name = data['name']
--- a/lib/discordrb/data/emoji.rb
+++ b/lib/discordrb/data/emoji.rb
@@ -18,7 +18,7 @@ module Discordrb
     attr_reader :animated
     alias_method :animated?, :animated
 
-    def initialize(data, bot, server)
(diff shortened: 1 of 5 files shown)


swarley:
This PR also includes a breaking change that removes the `bot` param from the `Emoji` constructor in its current state

lib/discordrb/data/emoji.rb
```diff
@@ -8,18 +8,17 @@ class Emoji
...
+    # @return [Array<Role>, nil] roles this emoji is active for
     attr_reader :roles
 
     # @return [true, false] if the emoji is animated
     attr_reader :animated
     alias_method :animated?, :animated
 
-    def initialize(data, bot, server)
-      @bot = bot
+    def initialize(data, server = nil)
```
z64: This shouldn't even be public. Don't care if you keep the default, but it should be hidden.
(We have quite a few visible private ctors in the lib that need this treatment)
```suggestion
    # @!visibility private
    def initialize(data, server)
```

lib/discordrb/data/emoji.rb
```diff
@@ -8,18 +8,17 @@ class Emoji
     # @return [String] the emoji name
     attr_reader :name
 
-    # @return [Server] the server of this emoji
+    # @return [Server,] the server of this emoji
```
z64: Pls

lib/discordrb/data/emoji.rb
```diff
@@ -8,18 +8,17 @@ class Emoji
...
     attr_reader :server
 
-    # @return [Array<Role>] roles this emoji is active for
+    # @return [Array<Role>, nil] roles this emoji is active for
```
z64: Best to document this condition
```suggestion
    # @return [Array<Role>, nil] roles this emoji is active for, or nil if the emoji's server is unknown
```

Daniel-Worrall:
This was merged (with changes) via shardlab/discordrb@18c7e71269fe09b526623f68d7d200a6ff3159f8 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #674, opened 2019-11-11
Update VoiceBot#channel on VOICE_STATE_UPDATE

### LINKED ISSUE #672: `VoiceBot#channel` is not updated when bot is moved to different channel
# Summary

`channel` attribute of `VoiceBot` is never changed in runtime, which can cause some errors when bot is working with voice channel it is currently connected to (checking amount of users for instance).

---

## How to reproduce

Simple command bot as an example:

```ruby
# frozen_string_literal: true

# This CommandBot is created to test several interactions with voice connections.

require 'discordrb'

#bot = Discordrb::Commands::CommandBot.new token: 'B0T.T0KEN.here', prefix: '!'
voice_connection = nil

bot.command :join do |event|
  channel = event.author.voice_channel
  if channel
    voice_connection = bot.voice_connect channel
    "Joined channel #{channel.name}"
  else
    "You are not in voice channel!"
  end
end

bot.command :channel do |event|
  if voice_connection
    "Chilling at #{voice_connection.channel.name}"
  else
    "Not in the voice channel"
  end
end

bot.command :leave do |event|
  if voice_connection
    voice_connection.destroy
    "Leaving"
  else
    "I'm not even here"
  end
end

bot.run
```

**Ruby version:**

ruby 2.5.5p157 (2019-03-15 revision 67260) [x64-mingw32]

**Discordrb version:**

master commit: 5f3dcc1a00bae37da1efe9d40f71835a6a18eebd

z64:
Fixes #672

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -781,6 +781,17 @@ module Discordrb
 
       server.update_voice_state(data)
 
+      existing_voice = @voices[server_id]
+      if user_id == @profile.id && existing_voice
+        new_channel_id = data['channel_id']
+        if new_channel_id
+          new_channel = channel(new_channel_id)
+          @voices[server_id].channel = new_channel
+        else
+          voice_destroy(server_id)
+        end
+      end
+
       old_channel_id
     end
 
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -786,7 +786,7 @@ module Discordrb
         new_channel_id = data['channel_id']
         if new_channel_id
           new_channel = channel(new_channel_id)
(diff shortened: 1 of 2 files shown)


lib/discordrb/bot.rb
```diff
@@ -781,6 +781,17 @@ def update_voice_state(data)
 
       server.update_voice_state(data)
 
+      existing_voice = @voices[server_id]
+      if user_id == @profile.id && existing_voice
+        new_channel_id = data['channel_id']
+        if new_channel_id
+          new_channel = channel(new_channel_id)
+          @voices[server_id].channel = new_channel
```
swarley: might as well use existing_voice here

Issue #675, opened 2019-11-12
`Bot#stop` need small changes due to #652


fizvlad:
# Summary

As mentioned in changelog after #652:
> - **(breaking change)** Removed `no_sync` argument in `Bot#stop` as part of a refactor that removes Ruby 2.3 compatibility ([#652](https://github.com/discordrb/discordrb/pull/652), thanks @PanisSupraOmnia)

However actual `Bot#stop` and its documentation wasn't changed at all.

**Ruby version:**

ruby 2.5.5p157 (2019-03-15 revision 67260) [x64-mingw32]

**Discordrb version:**

master commit 6a230c203f8676add9b01411b51592a9737afd55

z64:
Fixed with #676

PR #676, opened 2019-11-12
Removed no_sync argument from Bot#stop

fizvlad:
# Summary

Fix for #675 

Removed argument from `Bot#stop`. Changed documentation

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -257,7 +257,8 @@ module Discordrb
 
     # Stops the bot gracefully, disconnecting the websocket without immediately killing the thread. This means that
     # Discord is immediately aware of the closed connection and makes the bot appear offline instantly.
-    def stop
+    # @note This method no longer takes an argument as of 3.4.0 
+    def stop(_no_sync = nil)
       @gateway.stop
     end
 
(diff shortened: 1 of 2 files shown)


z64:
The argument needs to stay for a deprecation cycle, and the changelog needs to be updated to reflect that.

lib/discordrb/bot.rb
```diff
@@ -257,9 +257,8 @@ def join
 
     # Stops the bot gracefully, disconnecting the websocket without immediately killing the thread. This means that
     # Discord is immediately aware of the closed connection and makes the bot appear offline instantly.
-    # @param no_sync [true, false] Whether or not to disable use of synchronize in the close method. This should be true if called from a trap context.
-    def stop(no_sync = false)
-      @gateway.stop(no_sync)
+    def stop
```
swarley:
```suggestion
    # @note This method no longer takes an argument as of 3.4.0 
    def stop(_no_sync = nil)
```

fizvlad: Shoud we use `@deprecated` insterad of `@note`?

swarley: unfortunately YARD does not give us a way to deprecate arguments, and would list the entire method as deprecated

z64: Yeah. It doesn't play well with linters and IDEs on the ecosystem - they yell at you that the whole method will be removed. I did it before and people weren't happy. :smile_cat:

fizvlad:
> changelog needs to be updated to reflect that

I've removed `breaking change` prefix as this won't break 3.3.0 code.

PR #679, opened 2019-11-14
Add dnd? method for User

kaine119:
# Summary
Adds a `User#dnd?` method alongside the rest of the `status?` methods. Also adds documentation for each method

---
## Added
* `User#dnd?`
* Documentation for `User#online?`, `offline?`, `idle?` and `dnd?`

## Changed
* Add grouping around the methods above (feel free to revert this before squashing)

### DIFF
--- a/lib/discordrb/data/user.rb
+++ b/lib/discordrb/data/user.rb
@@ -166,7 +166,7 @@ module Discordrb
       @discriminator == Message::ZERO_DISCRIM
     end
 
-    %i[offline idle online].each do |e|
+    %i[offline idle online dnd].each do |e|
       define_method(e.to_s + '?') do
         @status.to_sym == e
       end
--- a/lib/discordrb/data/user.rb
+++ b/lib/discordrb/data/user.rb
@@ -166,6 +166,14 @@ module Discordrb
       @discriminator == Message::ZERO_DISCRIM
     end
 
+    # @!method offline?
+    #   @return [true, false] whether this user is offline.
+    # @!method idle?
+    #   @return [true, false] whether this user is idle.
+    # @!method online?
+    #   @return [true, false] whether this user is online.
+    # @!method dnd?
+    #   @return [true, false] whether this user is set to do not disturb.
     %i[offline idle online dnd].each do |e|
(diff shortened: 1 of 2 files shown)


lib/discordrb/data/user.rb
```diff
@@ -166,7 +166,18 @@ def webhook?
       @discriminator == Message::ZERO_DISCRIM
     end
 
-    %i[offline idle online].each do |e|
+    # @!group Statuses
```
swarley: I'm not a fan of the group here personally, but if we keep it the syntax needs a fixup
```suggestion
    # @!group Statuses

```

lib/discordrb/data/user.rb
```diff
@@ -166,7 +166,18 @@ def webhook?
...
+    # @!method offline?
+    #   @return [true, false] whether this user is offline.
+    # @!method idle?
+    #   @return [true, false] whether this user is idle.
+    # @!method online?
+    #   @return [true, false] whether this user is online.
+    # @!method dnd?
+    #   @return [true, false] whether this user is on Do Not Disturb.
+    # @!endgroup
```
swarley:
```suggestion

    # @!endgroup
```

If we do keep the group directive, the beginning and end must be in separate declarations for YARD to render it properly

lib/discordrb/data/user.rb
```diff
@@ -166,7 +166,18 @@ def webhook?
...
-    %i[offline idle online].each do |e|
+    # @!group Statuses
+    # @!method offline?
+    #   @return [true, false] whether this user is offline.
+    # @!method idle?
+    #   @return [true, false] whether this user is idle.
+    # @!method online?
+    #   @return [true, false] whether this user is online.
+    # @!method dnd?
+    #   @return [true, false] whether this user is on Do Not Disturb.
```
swarley: not sure how I feel about the pascal casing for `do not disturb`. maybe something like this?
```suggestion
    #   @return [true, false] whether this user is set to do not disturb.
```
I'm sure this is opinionated though

z64:
Thank you!

PR #683, opened 2019-11-19
Update webhook to include type field

swarley:
# Summary

Include `type` reader for webhooks

---

## Added
`Webhook#type`

## Changed
Return signature for `Webhook#token` is now `[String, nil]` as channel follower webhooks do not have an associated token.

### DIFF
--- a/lib/discordrb/data/webhook.rb
+++ b/lib/discordrb/data/webhook.rb
@@ -14,11 +14,14 @@ module Discordrb
     # @return [Server] the server that the webhook is currently connected to.
     attr_reader :server
 
-    # @return [String] the webhook's token.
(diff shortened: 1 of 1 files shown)


lib/discordrb/data/webhook.rb
```diff
@@ -14,12 +14,15 @@ class Webhook
     # @return [Server] the server that the webhook is currently connected to.
     attr_reader :server
 
-    # @return [String] the webhook's token.
+    # @return [String, nil] the webhook's token (for Incoming Webhooks)
```
z64:
```suggestion
    # @return [String, nil] the webhook's token, if this is an Incoming Webhook.
```

PR #689, opened 2019-12-28
Fix argument to Emoji.new in Server#add_emoji and #edit_emoji

GrapeColor:
# Summary

Fixed arguments to Emoji in add_emoji and edit_emoji.

## Fixed

Emoji # initialize requires three arguments, but only one was specified.
So I added the necessary arguments to Server#add_emoji and #edit_emoji.

### DIFF
--- a/lib/discordrb/data/server.rb
+++ b/lib/discordrb/data/server.rb
@@ -542,11 +542,12 @@ module Discordrb
     # @param image [String, #read] A base64 encoded string with the image data, or an object that responds to `#read`, such as `File`.
     # @param roles [Array<Role, String, Integer>] An array of roles, or role IDs to be whitelisted for this emoji.
     # @param reason [String] The reason the for the creation of this emoji.
+    # @param type [String] The type is the content type of the image.
     # @return [Emoji] The emoji that has been added.
-    def add_emoji(name, image, roles = [], reason: nil)
+    def add_emoji(name, image, roles = [], reason: nil, type: 'jpg')
       image_string = image
       if image.respond_to? :read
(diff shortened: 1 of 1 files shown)


GrapeColor:
sorry.
Originally there was no intention to push to this branch.

This commit adds a type argument to add_emoji so that any content type can be specified.

If you have any problems, close the PR.

really sorry.

z64:
No worries. It's a small change, fine to fit in this PR.

GrapeColor:
Could this PR conflict with # 671 ?
Given that, you may only need to add a type argument.

z64:
Yeah, #671's changeset is better to apply first.

As for `type`, I wonder if it can be made more ergonomic for the most common use case.
Most users are probably going to do something like:
```rb
server.add_emoji("foo", File.open("foo.jpg"))
```
Which `add_emoji` can detect in a similar way to `#read`: (Not tested)
```rb
def add_emoji(name, image, roles = [], reason: nil, type: nil)
  if image.respond_to? :read
    # If the user didn't specify a type, and the `image` knows a path, 
    # try to get its extension:
    if type.nil? && image.respond_to? :path
      type = File.extname(image.path)[1..]
    end
    image_string = "data:image/#{type || 'jpg'};base64,"
    image_string += Base64.strict_encode64(image.read)
  end
  # ..
end
```
So then, `type` just exists to override the mime type, or to provide it if the image file has no extension:
```rb
server.add_emoji("foo", File.open("foo"), type: "png")
```

GrapeColor:
Close to integrate with #710.

PR #698, opened 2020-01-16
Intents

swarley:
# Summary

This adds support for the new intent system. This is a rough first pass for people to test out and review.

```ruby
require 'discordrb'

bot = Discordrb::Commands::CommandBot.new(token: ENV['BOT_TOKEN'], prefix: '.', intents: [:guild_messages])

bot.command(:hello) do |event|
  "hello world!"
end

bot.run
```

This bot would respond to commands in servers but not direct messages for example.

A list of intents can be found under `Discordrb::INTENTS`, or [here](https://github.com/discordapp/discord-api-docs/pull/1307).

Currently there is an issue with the default intent value (all intents) which is also being questioned in the comments of that pull request.

## Added
`Discordrb::INTENTS` - Hash of intent names and their respective values.

## Changed
`Discordrb::Bot#initialize` - add intents kwarg
`Discordrb::Commands::CommandBot#initialize` - add intents kwarg
`Discordrb::Gateway#initialize` - add intents positional

Change to the way that the ready event is fired, since the existing method will not be triggered until the first dispatch if you do not have the `guilds` intent.

### DIFF
--- a/lib/discordrb.rb
+++ b/lib/discordrb.rb
@@ -15,6 +15,78 @@ module Discordrb
   # The unix timestamp Discord IDs are based on
   DISCORD_EPOCH = 1_420_070_400_000
 
+  INTENTS = {
+    guilds: 1 << 0,
+    guild_create: 1 << 0,
+    guild_delete: 1 << 0,
+    guild_role_create: 1 << 0,
+    guild_role_update: 1 << 0,
+    guild_role_delete: 1  << 0,
+    channel_create: 1  << 0,
+    channel_update: 1  << 0,
+    channel_delete: 1 << 0,
+    channel_pins_update: 1  << 0,
+
+    guild_members: 1 << 1,
+    guild_member_add: 1 << 1,
+    guild_member_update: 1 << 1,
+    guild_member_remove: 1 << 1,
+
+    guild_bans: 1 << 2,
+    guild_ban_add: 1 << 2,
+    guild_ban_remove: 1 << 2,
+
+    guild_emojis: 1 << 3,        
+    guild_emojis_update: 1 << 3,
+
+    guild_integrations: 1 << 4,
+    guild_integrations_update: 1 << 4,
+
+    guild_webhooks: 1 << 5,
+    webhooks_update: 1 << 5,
+
+    guild_invites: 1 << 6,
+    invite_create: 1 << 6,
+    invite_delete: 1 << 6,
+
+    guild_voice_states: 1 << 7,
+    voice_state_update: 1 << 7,
+
+    guild_presences: 1 << 8,
+    presence_update: 1 << 8,
+
+    guild_messages: 1 << 9,
+    message_create: 1 << 9,
+    message_update: 1 << 9,
+    message_delete: 1 << 9,
+
+    guild_message_reactions: 1 << 10,
+    message_reaction_add: 1 << 10,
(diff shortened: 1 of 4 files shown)


lib/discordrb.rb
```diff
@@ -15,6 +15,77 @@ module Discordrb
   # The unix timestamp Discord IDs are based on
   DISCORD_EPOCH = 1_420_070_400_000
 
+  INTENTS = {
+    servers: 1 << 0,
+    server_create: 1 << 0,
```
z64: All these alias values need to be removed; Intents can't actually be configured like this, so it is just confusing as to why they exist / who would use these values..

lib/discordrb/bot.rb
```diff
@@ -1397,5 +1400,24 @@ def handle_awaits(event)
...
+      intents.reduce(0) do |sum, intent|
+        case intent
+        when Symbol
+          if INTENTS[intent]
+            sum | INTENTS[intent]
+          else
+            LOGGER.warn("Unknown intent: #{intent}")
+            sum
+          end
+        when Number
```
z64: I actually meant to just accept a packed integer at the front, but I kind of like this too. Probably better for inserting individual ad-hoc values.

jcraigk:
The list of intents included seems out of date here and the PR of the Discord API docs referenced in the description is closed.  Is this PR still alive?

swarley:
The lis of intents are technically up to date the issue lies in the fact that discordrb uses `Server` instead of `Guild. I've asked for feedback on reverting the change, as I feel as though if we are going to rely on external documentation we should have parity with the terms.

Secondly, this PR is still alive. I'm keeping an eye on any of my open PRs for review or updates in features.

jcraigk:
I also support changing `Server` to `Guild`.  Would warrant a new major version release.

As for list of intents, I was expecting to see the granular ones listed at https://discordapp.com/developers/docs/topics/gateway#gateway-intents (including `GUILD_MEMBER_ADD`, for example) but maybe I'm misunderstanding and only the top-level ones listed there are valid values to send during gateway connection (in this case `GUILD_MEMBERS`).

z64:
Correct, only the "top level" ones are the values that can be passed to the bitmask.
For the library as a whole, we won't be changing `Server` to `Guild` anytime soon, for many reasons. Maybe late 4.x or 5.x, but there is way more friction in doing so for very little gain, mostly because the library is so far abstract from how the API operates.

As for intents, the nomenclature will probably be stabilized in a later PR. The whole intents business is was a large change for Discord, and the dust is starting to settle - which is what I was waiting for. I think we will probably use `server_` naming, and write out our own documentation for this, expressed in terms of discordrb's API.

z64:
Moreover, adding compatibility for guild-* naming is drastically easier later, than it is to start thinking about guild-* now.

xokocodo:
I understand the naming is potentially confusing, but is that the only blocker for this PR?

Daniel-Worrall:
This was merged via shardlab/discordrb@639df583538cf89f437d7221802b1f575ae79428 and shardlab/discordrb@0c5c8501f733542b42f1e718ae619ea75eae9a75 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #699, opened 2020-01-22
typos

Daniel-Worrall:
# Summary

Couple redundant code removals.

Various typo fixups through spell-checker.

### DIFF
--- a/lib/discordrb/bot.rb
+++ b/lib/discordrb/bot.rb
@@ -168,10 +168,7 @@ module Discordrb
     #   The list of emoji the bot can use.
     #   @return [Array<Emoji>] the emoji available.
     def emoji(id = nil)
-      gateway_check
-      unavailable_servers_check
-
-      emoji_hash = @servers.values.map(&:emoji).reduce(&:merge)
+      emoji_hash = servers.values.map(&:emoji).reduce(&:merge)
       if id
         id = id.resolve_id
         emoji_hash[id]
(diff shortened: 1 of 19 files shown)


lib/discordrb/data/server.rb
```diff
@@ -23,7 +23,7 @@ class Server
     include IDObject
     include ServerAttributes
 
-    # @return [String] the ID of the region the server is on (e.g. `amsterdam`).
+    # @return [String] the ID of the region the server is on (e.g. `Amsterdam`).
```
swarley: `amsterdam` is actually correct in this context. `region_id` differs from the region name

lib/discordrb/bot.rb
```diff
@@ -168,10 +168,7 @@ def servers
     #   The list of emoji the bot can use.
     #   @return [Array<Emoji>] the emoji available.
     def emoji(id = nil)
-      gateway_check
-      unavailable_servers_check
```
z64: This check needs to stay

Daniel-Worrall: This check is called in Bot#servers

z64: Ah, right. NP then

Daniel-Worrall:
Merged in the [new repo](https://github.com/shardlab/discordrb) in cc2900b582f4a7141318833fa2f5e7d754e1b6f4

Issue #700, opened 2020-02-14
The calculation result of PermissionCalculator#permission? is incorrect


GrapeColor:
# Summary
The calculation result of Discordrb::PermissionCalculator#permission? is incorrect.
According to the Discord specification, if there are multiple roles, "Allow" always takes precedence when overwriting channel permissions, regardless of the priority of the roles.
However, for "permission?", the authority of the upper role takes precedence, and the "Allow" of the lower role is ignored.
Therefore, authority cannot be checked correctly.
It is not possible to identify where the problem is in the library.

---

## Environment

**Ruby version:**

ruby 2.6.5p114 (2019-10-01 revision 67812) [x64-mingw32]

**Discordrb version:**

C:/Ruby26-x64/lib/ruby/gems/2.6.0/bundler/gems/discordrb-2c2b50e32c05

GrapeColor:
# Channel permission setting
![image](https://user-images.githubusercontent.com/40069549/74504871-efe8dd80-4f38-11ea-9095-b928853a164a.png)
![image](https://user-images.githubusercontent.com/40069549/74504880-faa37280-4f38-11ea-9ec5-9c359c22c1e1.png)

# Testing code
```ruby
require 'bundler/setup'
require 'discordrb'

bot = Discordrb::Commands::CommandBot.new(
  token: "TOKEN",
  prefix: "/"
)


bot.command(:test) do |event|
  p event.server.member(bot.profile.id).permission?(:read_messages, event.channel)
end

bot.run
```

# Result
```ruby
false
```

swarley:
Sorry that it took a while to get to this one, it appears we discussed this in the channel but I forgot to commit something afterwards.

GrapeColor:
No problem.
Thank you for your discussion and problem solving.🍰

GrapeColor:
I'm sorry.
It seems that there are still problems.
For example, when two roles are given, overwriting only one of them seems to make a wrong decision.

Specific example:

## Server roles
![image](https://user-images.githubusercontent.com/40069549/80872716-b1b8ba00-8cee-11ea-8376-84c15dc06b90.png)

## Channel overwrites
![image](https://user-images.githubusercontent.com/40069549/80872756-f7758280-8cee-11ea-9d7a-43781cdf2958.png)

## Testing code
```ruby
require 'bundler/setup'
require 'discordrb'

bot = Discordrb::Commands::CommandBot.new(token: 'TOKEN', prefix: '/')

[...]

PR #701, opened 2020-02-14
Fix PermissionCalculator

aaaa777:
# Summary
Once an action allowed by any role, action is always allowed.

Official permission hierarchy is below.
https://support.discordapp.com/hc/articles/206141927-How-is-the-permission-hierarchy-structured-


aaaa777:
See also #700

aaaa777:
I created pull request before visit discord channel...
Now this request is no longer needed :(
This is my first pull request............

PR #703, opened 2020-02-17
Refactor eval.rb

gsbhasin123:
To make it more useful :p
I'ma make as many PRs as possible


gsbhasin123:
Realized this is useless

PR #712, opened 2020-04-23
Quick success on permission calculation from allow overwrites

### LINKED ISSUE #700: The calculation result of PermissionCalculator#permission? is incorrect

swarley:
Closes #700 

End calculation early when an `allow` override is met

---

## Fixed
`PermissionCalculator#defined_role_permission?`

### DIFF
--- a/lib/discordrb/permissions.rb
+++ b/lib/discordrb/permissions.rb
@@ -187,15 +187,13 @@ module Discordrb
       roles_to_check.sort_by(&:position).reduce(false) do |can_act, role|
         # Get the override defined for the role on the channel
         channel_allow = permission_overwrite(action, channel, role.id)
-        can_act = if channel_allow
-                    # If the channel has an override, check whether it is an allow - if yes,
-                    # the user can act, if not, it can't
-                    channel_allow == :allow
-                  else
-                    # Otherwise defer to the role
-                    role.permissions.instance_variable_get("@#{action}") || can_act
-                  end
-        can_act
+        
+        # If the channel has an override, check whether it is an allow - if yes,
+        # the user can act, if not, it can't
+        return true if channel_allow == :allow
+
+        # Otherwise defer to the role
+        role.permissions.instance_variable_get("@#{action}") || can_act
       end
     end
 


lib/discordrb/permissions.rb
```diff
@@ -190,10 +190,14 @@ def defined_role_permission?(action, channel)
 
         # If the channel has an override, check whether it is an allow - if yes,
         # the user can act, if not, it can't
-        return true if channel_allow == :allow
+        if channel_allow
```
GrapeColor, line 193: If `channel_allow` is `false`, this condition will check role permissions.
For example, how about `channel_allow != nil`?

GrapeColor:
What about setting the `overwrited` flag so that if there is a channel overwrite, that one has priority?
```ruby
overwrited = false  # add
roles_to_check.sort_by(&:position).reduce(false) do |can_act, role|
  # Get the override defined for the role on the channel
  channel_allow = permission_overwrite(action, channel, role.id)

  # If the channel has an override, check whether it is an allow - if yes,
  # the user can act, if not, it can't
  if channel_allow
    overwrited = true  # add
    return true if channel_allow == :allow

    false
  else
    # Otherwise defer to the role
    !overwrited && role.permissions.instance_variable_get("@#{action}") || can_act  # add
  end
end
```

Daniel-Worrall:
This was merged via shardlab/discordrb@9cc3a26c5f967b87745e20ad1ebc71e761973650 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #715, opened 2020-04-28
Properly serialize unicode emojis in `Emoji#mention`

swarley:
# Summary

Allow `Emoji#mention` to be used for unicode emojis.

---

## Added
Tests for serialization of unicode emojis via `mention`

## Fixed
`Emoji#mention` will return `#name` if `#id` is nil, where it would previously return `<:name:>`

### DIFF
--- a/lib/discordrb/data/emoji.rb
+++ b/lib/discordrb/data/emoji.rb
@@ -42,6 +42,8 @@ module Discordrb
 
     # @return [String] the layout to mention it (or have it used) in a message
     def mention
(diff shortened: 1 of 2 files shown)


Daniel-Worrall:
This was merged via shardlab/discordrb@0a32817190e0ce649e59d44722aa4114be0c2a25 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #716, opened 2020-04-29
fix: Discordrb::Bot/Cache.server loads a server's channels.

schwern:
# Summary

`Bot.server(id)` does not load a server's channels. While `Server.new` expects channels to be in the data. `Server#process_channels` is private leaving the Bot with no obvious way to set the channels.

Perhaps an older version of the API did send the channels when fetching a server. Now they must be fetched separately. Or perhaps I'm using Bot wrong?

## Changed

`Cache#server` now loads a server's channels.

### DIFF
--- a/lib/discordrb/cache.rb
+++ b/lib/discordrb/cache.rb
@@ -96,11 +96,16 @@ module Discordrb
 
       LOGGER.out("Resolving server #{id}")
       begin
-        response = API::Server.resolve(token, id)
+        data = JSON.parse(
+          API::Server.resolve(token, id)
+        )
+        data['channels'] = JSON.parse(
+          API::Server.channels(token, id)
+        )
       rescue Discordrb::Errors::NoPermission
         return nil
       end
-      server = Server.new(JSON.parse(response), self)
+      server = Server.new(data, self)
       @servers[id] = server
     end
 


swarley:
If you're not connected to the gateway (via `Bot#run`) you lose out on certain cache data, as channels are present in the [`GUILD_CREATE`](https://discordapp.com/developers/docs/topics/gateway#guild-create) payload

schwern:
@swarley `Bot#servers` is being used in `Bot#ready`. We do call `Bot#run`.

See https://github.com/schwern/conexus/blob/fixups/bot.rb#L55

z64:
If we want this, the proper way to do it is lazily via `Server#channels`. Eagerly making two API requests on a cache miss isn't ideal.
But yes - I'd question whether this is needed at all in the first place, especially if `Bot` is being used correctly as mentioned, etc..

schwern:
@z64 Is https://github.com/schwern/conexus/blob/fixups/bot.rb#L55 using Bot incorrectly?

schwern:
While I agree this should be done lazily, that's a pretty big redesign.

How about if-and-only-if the channels are not present in the data?

```
        data['channels'] ||= JSON.parse(
          API::Server.channels(token, id)
        )
```

z64:
> While I agree this should be done lazily, that's a pretty big redesign.

To be clear, we already do this sort of memoization for other `Server` properties, and for other data types as well:

https://github.com/discordrb/discordrb/blob/master/lib/discordrb/data/server.rb#L364-L367

However, there are specific reasons we do these - these fields may not be filled on the initial `GUILD_CREATE` payload that we receive. There is also the lazy `Server#owner` method, a behavior which was introduced to reduce HTTP load under certain bot configurations. There's also the `#widget_` methods, etc.

Making a similar lazy `Server#channels` method would be the correct approach for our established style. However, there must be *some* reason to do so, as stated above. All guild channels are sent with `GUILD_CREATE`, which we then cache; there shouldn't be a normal scenario where we don't have this data while a session is running. Using the bots cache *without* a running session is something we don't currently support, and issue warnings or errors where practical if you try to do so. ([Disclaimer in our wiki about this](https://github.com/discordrb/discordrb/wiki/API-Wrapper#cant-i-just-use-a-bot-without-connecting))

---

Back on topic: I'm fine with making this change, but we have gone - as far as I am aware - discordrb's entire life without needing to do this. So I would like to know under what conditions this is happening - it could be some other bug entirely. Any other incident of this occurring has been from people using `Bot` without an active session as described, and was fixed by appropriately accessing data inside event handlers (as opposed to outside of them).

I personally haven't had time to closely audit your code yet beyond a brief skim - nothing jumped out at me. I will take a closer look at some point, but it would be helpful if you can dial in a reproducible scenario. You can enable logging of the raw data passing through the library with `log_mode: :debug` in `Bot.new`.

Of course, if you need detailed assistance with doing so, feel free to join our [library channel](https://discord.gg/cyK3Hjm) where myself or others can work more closely with you to investigate this.

z64:
It was also pointed out that your lockfile requires discordrb 3.2.1, which is quite old at this point. I'm assuming this lockfile is up to date, of course..

While I'm not aware of any particular issue with 3.2.1 that would affect your scenario, I strongly recommend upgrading.

schwern:
@z64 [Here is a stripped down program demonstrating the problem](https://gist.github.com/schwern/356691c2df5b3a1bd7888d09533977dd). It should connect and list all the voice and text channels. It connects, but lists nothing.

Yes, I'd already done a bundle update and discordrb is at 3.3.0. I hadn't pushed it. This is someone else's project I picked up and found this problem trying to get it to work.

I'll look into doing channels on demand.

z64:
With that sample, I can't reproduce this on any of my bots.

schwern:
@z64 Could it be a server setting? Perhaps "Intents", which I don't understand.

z64:
I can actually reproduce this, but I'm not sure the cause. I'll look into it further.

z64:
Ok, I see the issue. The problem is guild-related events may be received before `GUILD_CREATE`, causing the guild to be fetched over HTTP. The impending `GUILD_CREATE` data will then get discarded.

z64:
I have a patch that can adjust the semantics of our `ensure` cache methods to make us resilient to this happening, but honestly, I'm unsure if it's normal API behavior in the first place. This seems to go against some "unspoken guarantees". The patch should be good on its own, so I'll publish it either way, but I'll try to dig up more background on this behavior..

As imagined, this PR's changeset won't be necessary - but thanks for reporting, and for your interest in contributing. This is an odd one.

z64:
I would also say that issuing the request to update the bot's username before connecting is a likely catalyst in this. While I'm unable to rule out it still might happen, this is probably a bad idea to do this outside of a `bot.ready` handler.

schwern:
An API is behaving abnormally?!

Tommy Lee Jones - This is my shocked face

(I work with many, many APIs as my day job. Situation normal.)

Thanks for digging into it.

schwern:
@z64 [I've stripped down the example even further](https://gist.github.com/schwern/356691c2df5b3a1bd7888d09533977dd) removing the calls outside the ready handler and using synchronous run. Sometimes it loads the channels, sometimes it does not.

Definitely seems like a timing issue.

z64:
You have to be careful when running that code as well. Because you do not cleanly exit, the gateway may still report an active (hanging) session from your previous attempt, causing the same issue despite improved code. Make sure the bot shows as offline before trying.

The main culprit here, is that there was recently a compatibility change made to the API to propagate `USER_UPDATE` events (which are raised for *your bots* own account, and no one else's) into `GUILD_MEMBER_UPDATE` events. The former will not fire until all guilds have been sent, but the latter is missing this behavior.. which sounds like an oversight given the context of the change.

schwern:
@z64 What is the recommended bot exit procedure? I don't see anything special in the [shutdown example](https://github.com/discordrb/discordrb/blob/master/examples/shutdown.rb). Nor is it mentioned in [Basic Bot Creation](https://github.com/discordrb/discordrb/wiki/Basic-bot-creation).

z64:
`Bot#stop`

ishowta:
@z64 
> I have a patch that can adjust the semantics of our `ensure` cache methods to make us resilient to this happening, but honestly, I'm unsure if it's normal API behavior in the first place. This seems to go against some "unspoken guarantees". The patch should be good on its own, so I'll publish it either way, but I'll try to dig up more background on this behavior..

Do you plan to publish the patch?

PR #717, opened 2020-04-29
Change reconnection close code to 4000

GrapeColor:
# Summary
As said in discord/discord-api-docs#1472, reconnecting with a close code 1000 seems to be a problem.
Therefore, referring to [Rapptz/discord.py 403651a](https://github.com/Rapptz/discord.py/commit/403651a1440260cb46f42d68a82ee35e193a350d), I changed it to set the close code 4000 when reconnecting.

## Changed
Changed to be able to set `code` for close frame in `Gateway#send` and `Gateway#close`.
Changed to set close code 4000 in `Gateway#reconnect`.

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -367,7 +367,7 @@ module Discordrb
       @instant_reconnect = true
       @should_reconnect = true
 
-      close
+      close code: 4000
     end
 
     # Sends a resume packet (op 6). This replays all events from a previous point specified by its packet sequence. This
@@ -808,7 +808,7 @@ module Discordrb
       end
     end
 
-    def send(data, type = :text)
+    def send(data, type = :text, code = nil)
       LOGGER.out(data)
 
       unless @handshaked && !@closed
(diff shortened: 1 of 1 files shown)


z64:
I'm fine with making some changes to this, however I think the right approach for now is:
- Default to 4000 for internal closures.
- `bot.stop` should flow down `1000`, since we don't support resuming this way yet.

The reasoning:

This isn't *exactly* a problem for us. We should send a non-1000 close code for a *user-initiated* close (`bot.stop`), but the significance of sending a non-1000 code is that it allows the session to be resumed later - but discordrb actually provides no way to resume a previous session like this. You would need to have saved the `session_id` somewhere, and then somehow pass that back to `bot`.. there is no method to do this.

When *discord* initiates a reconnect, i.e. with Op 9 Invalid Session, Op 7 Reconnect, or other unannounced closure - discordrb *will* already attempt and successfully resume on these.

Discord currently doesn't have any guidelines except a "non-1000" close code. So I'm not sure if 4000-4999 range close code is appropriate. But this code is fine until that thread receives some clarification. 

---

References:
- [IANA Close Code registry](https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number)
- [RFC 6455 7.4.1](https://tools.ietf.org/html/rfc6455#section-7.4.1)

GrapeColor:
Thank you for your detailed review.

I seem to have misunderstood the logic of Gateway in discordrb.(Especially about sessions)
I can't really change the Gateway logic here, so I'll close the PR.

Thank you for taking the time.

PR #719, opened 2020-04-30
docs: Demonstrate proper bot shutdown.

schwern:
This updates all the examples to demonstrate proper shutdown using
`bot.stop` when the process is killed. Or, in the case of the shutdown
example, before exiting.

# Summary

Demonstrate proper bot shutdown procedure in the examples.

@z64 [told me I was not correctly shutting down my bot](https://github.com/discordrb/discordrb/pull/716#issuecomment-621525294), but I'd followed the examples and [wiki](https://github.com/discordrb/discordrb/wiki/Basic-bot-creation) (which has been updated).

## Changed

Updated all the examples to demonstrate proper shutdown using `bot.stop` when the process is killed. Or, in the case of the shutdown example, before exiting.

### DIFF
--- a/examples/shutdown.rb
+++ b/examples/shutdown.rb
@@ -15,7 +15,14 @@ bot.command(:exit, help_available: false) do |event|
   break unless event.user.id == 66237334693085184 # Replace number with your ID
 
   bot.send_message(event.channel.id, 'Bot is shutting down')
+  bot.stop
   exit
 end
 
-bot.run
+# Connect to Discord
+begin
+  bot.run
+# Disconnect when killed with ctrl-c
+rescue Interrupt
+  bot.stop
+end
--- a/examples/awaits.rb
+++ b/examples/awaits.rb
@@ -76,7 +76,12 @@ bot.message(content: '!time') do |event|
 end
 
 # Connect to Discord
-bot.run
+begin
+  bot.run
+# Disconnect when killed with ctrl-c
+rescue Interrupt
+  bot.stop
+end
 
 # For more details about Awaits, see:
 # https://www.rubydoc.info/gems/discordrb/Discordrb/Await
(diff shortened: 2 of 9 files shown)


z64:
Thanks for this!

I am unsure though that this is the best example. We do our best to make the library approachable for beginners, and `begin..rescue Interrupt` I feel might leave a lot of questions. What I think might be a better approach is one of these:

- Keep `examples/` focused on discordrb code: Make a `!stop` command that issues `bot.stop`.
```rb
bot.message(from: ..example ID here.., content: "!stop) do |event|
  event.respond("bye!")
  bot.stop
end
```
- I *believe* an `at_exit` handler will also work. This might be preferred over `begin..rescue Interrupt`. I'm not picky about a one line `{ ... }` over `do..end` here, I think one line is enough though.
```rb
at_exit { bot.stop }
bot.run
```

For the Wiki, that can be a space to discuss alternative means of exiting - `begin..rescue Interrupt`, `Signal.trap("INT")`, with a brief explanation about signals for those reading the Wiki. I'd rather explain that SIGINT is sent when you press CTRL+C, and not have the examples depend on you having read the Wiki to be clued into that.

WDYT?

schwern:
I didn't know about `at_exit`. Yes, that is far better.

I do believe that all the examples should demonstrate proper basic bot procedure. `at_exit { bot.stop }` is less intrusive.

examples/shutdown.rb
```diff
@@ -15,7 +15,11 @@
   break unless event.user.id == 66237334693085184 # Replace number with your ID
 
   bot.send_message(event.channel.id, 'Bot is shutting down')
+  bot.stop
   exit
```
z64, line 19: We can omit this line now. After this, execution will continue following `bot.run` since it's not blocking anymore, and the script will exit normally.

PR #721, opened 2020-05-13
Support :help_message attribute in CommandBot

jubishop:
Allow customization of `!help` command text.

Users report they didn't realize after just typing `!help` that they could get more info by typing `!help [command]`

I've added the attribute `:help_message` to the constructor of CommandBot to customize the output of the text when a user enters `!help` (or equivalent help command). 

I use this in my code by passing `help_message: "**List of commands:**\n  *(Type !help [command] for more info)*"`

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -27,7 +27,6 @@ module Discordrb::Commands
     # @see Discordrb::Bot#initialize Discordrb::Bot#initialize for other attributes that should be used to create the underlying regular bot.
     # @option attributes [String, Array<String>, #call] :prefix The prefix that should trigger this bot's commands. It
     #   can be:
-    #
     #   * Any string (including the empty string). This has the effect that if a message starts with the prefix, the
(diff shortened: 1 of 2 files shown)


lib/discordrb/commands/command_bot.rb
```diff
@@ -44,6 +43,7 @@ class CommandBot < Discordrb::Bot
     # @option attributes [Symbol, Array<Symbol>, false] :help_command The name of the command that displays info for
     #   other commands. Use an array if you want to have aliases. Default is "help". If none should be created, use
     #   `false` as the value.
+    # @option attributes [String] :help_message The message that will display above the list of commands.  Default is "**List of commands:**".  If :help_command is set to false, this attribute will be ignored.
```
greenbigfrog, line 46: Line should wrap

jubishop: yeah this is weird because in my checkout none of these lines wrap at 80chars.  don't understand why it changes upon submission.

jubishop: oh nvm they do wrap it just happens to be at 120 and not 80.  ok updating.

PR #727, opened 2020-06-11
Close code for reconnect

swarley:
# Summary

Add a close code argument to `Gateway#close` for use during reconnect to preserve our session on resume. 

`websocket-ruby` [does not support sending a 1012 (Service Restart)](https://github.com/imanel/websocket-ruby/blob/master/lib/websocket/frame/handler/handler07.rb#L44) so a 4000 close code was selected.
Surveying other libraries it seems they reconnect using 4000 as well.

## Changed
`Gateway#close` - Added optional `code` argument that defaults to 1000
`Gateway#handle_reconnect` - Now calls `close` with a 4000 close code
`Gateway#send` - Added optional `code` argument for use with close codes

### DIFF
--- a/lib/discordrb/gateway.rb
+++ b/lib/discordrb/gateway.rb
@@ -367,7 +367,7 @@ module Discordrb
       @instant_reconnect = true
       @should_reconnect = true
 
-      close
+      close(4000)
     end
 
     # Sends a resume packet (op 6). This replays all events from a previous point specified by its packet sequence. This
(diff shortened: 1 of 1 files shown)


Daniel-Worrall:
This was merged (with changes) via shardlab/discordrb@018e401cca7590eabfa325ef76848949af7d9db8 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #730, opened 2020-06-21
Fix type of ServerMemberDeleteEvent#user

sarna:
# Summary
`ServerMemberDeleteEvent#user` is inherited from `ServerMemberEvent` - there it says it returns a `Member`, which is wrong in the case of `ServerMemberDeleteEvent`: the user is not a `Member` anymore, they left the server. I tried calling `Member` methods on it in my bot and was getting confused why they wouldn't work - I think this PR would clear up the confusion for other people. 

---

## Added
Documentation for `ServerMemberDeleteEvent#user` - now it correctly says it returns a `User`.

### DIFF
--- a/lib/discordrb/events/members.rb
+++ b/lib/discordrb/events/members.rb
@@ -84,11 +84,8 @@ module Discordrb::Events
       @user = Discordrb::User.new(data['user'], bot)
     end
 
-    # @!attribute [r] user
-    #   @return [User] the user in question.
-    def user
-      super
-    end
+   # @return [User] the user in question.
+   attr_reader :user
   end
 
   # Event handler for {ServerMemberDeleteEvent}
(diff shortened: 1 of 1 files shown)


lib/discordrb/events/members.rb
```diff
@@ -83,6 +83,12 @@ class ServerMemberDeleteEvent < ServerMemberEvent
     def init_user(data, bot)
       @user = Discordrb::User.new(data['user'], bot)
     end
+
+    # @!attribute [r] user
+    #   @return [User] the user in question.
+    def user
+      super
+    end
```
swarley, line 91:
```suggestion
   # @return [User] the user in question.
   attr_reader :user
```
This is the format z64 requested

Daniel-Worrall:
This was merged (with changes) via shardlab/discordrb@648596335da886b01ed87ff6f43c8bc2a7e0aae4 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #732, opened 2020-07-09
Use KILL signal on windows

swarley:
# Summary

Windows does not support using `TERM` as a kill signal. This changes to using `KILL` on Windows platforms and verifies that our child process has not already exited before attempting a `Process.kill`.

---

## Changed
`VoiceBot#play` - Signal used for `Process.kill` is now platform dependent.


Daniel-Worrall:
This was merged via shardlab/discordrb@0e26a418836e211586b1a162f6ca575ec5f6d7aa in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #743, opened 2020-10-19
Add VoiceServerUpdateEvent

swarley:
# Summary

Expose a `VoiceServerUpdateEvent` that can be used for libraries like [telecom](https://github.com/b1naryth1ef/telecom) and [lavalink](https://github.com/Frederikam/Lavalink).

Previously the event handler was used only internally and had a stub saying that it was not useful for end users, but with an external audio player use case this is not true.

---

## Added
`Events::VoiceServerUpdateEvent` - Event raised by the bot.
`Events::VoiceServerUpdateEventHandler` - Handler created from the container method.
`Container#voice_server_update` - Method to hook into a `VOICE_SERVER_UPDATE` event.

### DIFF
--- a/lib/discordrb/events/voice_server_update.rb
+++ b/lib/discordrb/events/voice_server_update.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'discordrb/events/generic'
+require 'discordrb/data'
+
+module Discordrb::Events
+  # Event raised when a server's voice server is updating.
+  # Sent when initially connecting to voice and when a voice instance fails
+  # over to a new server.
+  # This event is exposed for use with library agnostic interfaces like telecom and
+  # lavalink.
+  class VoiceServerUpdateEvent < Event
+    # @return [String] The voice connection token
+    attr_reader :token
+
+    # @return [Server] The server this update is for.
+    attr_reader :server
+
+    # @return [String] The voice server host.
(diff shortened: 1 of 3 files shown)


lib/discordrb/events/voice_server_update.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      [
+        matches_all(@attributes[:from], event.server) do |a, e|
+          a == if a.is_a? String
+                 e.name
+               else
+                 e
+               end
+        end
+      ]
```
nut4k1, line 44: for me this looks prettier
```
[matches_all(@attributes[:from], event.server) { |a, e| a == a.is_a?(String) ? e.name : e }]
```

swarley: It's formatted in this way to keep consistency with other handlers, otherwise I would condense it. 
Personally I wouldn't allow for matching the name at all if I had a choice, but that's the behavior for `VoiceStateUpdateHandler` so I left it as is.

Daniel-Worrall:
This was merged via shardlab/discordrb@f0701d22d1dea8bdbd8ed48262ca3c2ca23e47a6 in the [new repo](https://github.com/shardlab/discordrb) and released in [v3.4.0](https://github.com/shardlab/discordrb/releases/tag/v3.4.0) and can be closed.

PR #750, opened 2021-01-24
Allow `CommandBot` `command_doesnt_exist_message` to be a callable

kmcphillips:
# Summary

Allows the `CommandBot` `command_doesnt_exist_message:` to be a callable. The behaviour stays the same for existing strings and `nil` passed in, but if instead something that responds to `:call` is passed in, that is first called with the `event`, then the result handled in the same way as above. This includes the `nil` behaviour and the replacing of the `"%command%"` substring.

Example:

```ruby
bot = Discordrb::Commands::CommandBot.new(
  token: token,
  prefix: "!",
  command_doesnt_exist_message: ->(event) {
    "Your command %command% doesn't exist. Have you tried? #{ find_similar_commands(event) }"
  }
)
```

## Added

* `CommandBot` `command_doesnt_exist_message:` can be be a callable.

# Alternative approach?

The other approach I was considering was adding:
```ruby
class CommandBot
  def command_not_found(&block)
    @command_not_found = block
  end
end
```

It would be invoked in the same way when a command is missing, but it would mean that it would be defined like:

```ruby
bot = Discordrb::Commands::CommandBot.new(token: token)
bot.command_not_found do |event|
  "Your command %command% doesn't exist. Have you tried? #{ find_similar_commands(event) }"
end
```

This is more in line with the DSL to define commands and responses, but would mean deprecating the `command_doesnt_exist_message` argument, then bumping the semver version.

I would be happy to implement this approach instead. It's mostly working on a branch.

### DIFF
--- a/lib/discordrb/commands/command_bot.rb
+++ b/lib/discordrb/commands/command_bot.rb
@@ -44,9 +44,10 @@ module Discordrb::Commands
     # @option attributes [Symbol, Array<Symbol>, false] :help_command The name of the command that displays info for
     #   other commands. Use an array if you want to have aliases. Default is "help". If none should be created, use
     #   `false` as the value.
-    # @option attributes [String] :command_doesnt_exist_message The message that should be displayed if a user attempts
+    # @option attributes [String, #call] :command_doesnt_exist_message The message that should be displayed if a user attempts
     #   to use a command that does not exist. If none is specified, no message will be displayed. In the message, you
-    #   can use the string '%command%' that will be replaced with the name of the command.
+    #   can use the string '%command%' that will be replaced with the name of the command. Anything responding to call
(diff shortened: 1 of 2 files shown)


swarley:
Hello! Would you mind opening this on our new repo https://github.com/shardlab/discordrb? I'd also invite you to join our channel in discord api, https://discord.gg/cyK3Hjm, as we have plans for reworking commandbot for the next major release and would love input during design discussions.

kmcphillips:
Oh sure. Forked?

I'll head over there.

swarley:
Official continuation on a hard fork, you can see the repo listed on the latest gem release https://rubygems.org/gems/discordrb/versions/3.4.0

kmcphillips:
Great. Reopened as https://github.com/shardlab/discordrb/pull/25

